PR #77, opened 2026-02-26
Feature/issue 65 timing

### LINKED ISSUE #65: Add Pipeline Execution Timing Display
**Difficulty:** Medium
**Area:** full stack

## Description

When users build and execute image processing pipelines, they have no visibility into how long each operation takes or the total execution time. This information is valuable for several reasons: it helps users understand the computational cost of different operations, it aids in optimizing pipelines by identifying bottlenecks, and it provides useful feedback during longer-running pipelines to confirm that processing is actually happening.

This issue involves measuring execution time on the backend (per-step and total), returning timing data in the pipeline response, and displaying it on the frontend in a clear and unobtrusive way.

## Acceptance Criteria

- [ ] The pipeline API response includes total duration and per-step timing data
- [ ] Per-step timing includes the step index, operator type, and duration in milliseconds
- [ ] Each pipeline step is individually timed using a high-resolution timer
- [ ] Total pipeline duration is measured (including image decode/encode)
- [ ] Timing data is returned in the API response alongside the processed image
- [ ] The frontend data model is updated to handle timing fields from the API
- [ ] Total execution time is displayed in the UI after pipeline execution
- [ ] Per-step timing is available (in a collapsible detail view or similar)
- [ ] Timing is cleared when a new pipeline execution starts
- [ ] Timing display does not appear when no pipeline has been executed
- [ ] Existing API consumers are not broken (timing fields are optional/nullable)

## Resources

- [Python time.perf_counter()](https://docs.python.org/3/library/time.html#time.perf_counter) (high-resolution timer)
- [Pydantic models](https://docs.pydantic.dev/latest/)
- Pipeline executor: `imagelab-backend/app/services/pipeline_executor.py`
- Pipeline models: `imagelab-backend/app/models/pipeline.py`
- Frontend types: `imagelab-frontend/src/types/pipeline.ts`
- Zustand store: `imagelab-frontend/src/store/pipelineStore.ts`
- PreviewPane: `imagelab-frontend/src/components/Preview/PreviewPane.tsx`

Jayant-kernel:
## Description
This PR implements the pipeline action durations to surface per-operator execution times to the user. It restructures the backend to return a robust nested `PipelineTimings` model and updates the frontend Zustand store and UI to render these timings natively.
**Bonus: React ErrorBoundary** — This PR also wraps the editor and preview in a global error boundary that catches unexpected render crashes and shows a recovery UI.
**Advantages over naive implementation:**
Partial `step_timings` are returned even on pipeline failure. This means users can see exactly how far execution got and how long each completed step took before an error occurred. This fulfills the acceptance criteria requirement that "Partial timings are also properly recorded and surfaced if a step fails."
Fixes #65
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
## How Has This Been Tested?
Tested locally by spinning up both the PyTorch backend and Vite frontend, then manually executing pipelines with `Geometric` blocks to ensure accurate render timing and error reporting. 
- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 
issue65

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/models/pipeline.py
+++ b/imagelab-backend/app/models/pipeline.py
@@ -14,7 +14,7 @@ class PipelineRequest(BaseModel):
 
 class StepTiming(BaseModel):
     step: int
-    type: str
+    operator_type: str
     duration_ms: float
 
 
--- a/imagelab-backend/app/services/pipeline_executor.py
+++ b/imagelab-backend/app/services/pipeline_executor.py
@@ -8,6 +8,14 @@ NOOP_TYPES = {"basic_readimage", "basic_writeimage", "border_for_all", "border_e
 
 
 def execute_pipeline(request: PipelineRequest) -> PipelineResponse:
+    """
+    Execute the image-processing pipeline described by *request*.
+
+    Returns a PipelineResponse that always includes a ``timings`` field
+    populated with every step that completed before the function returned,
+    even when the response indicates failure.  This allows callers to
+    inspect partial execution progress on error.
+    """
     t_start_total = time.perf_counter()
     step_timings: list[StepTiming] = []
 
@@ -31,23 +39,25 @@ def execute_pipeline(request: PipelineRequest) -> PipelineResponse:
             t_fail = time.perf_counter()
             return PipelineResponse(
                 success=False,
-                error=f"Unknown operator '{step.type}' at step {i}",
-                step=i,
+                error=f"Unknown operator '{step.type}' at step {i + 1}",
+                step=i + 1,
                 timings=PipelineTimings(total_ms=(t_fail - t_start_total) * 1000, steps=step_timings),
             )
 
         try:
-            operator = operator_cls(step.params)
             t_step_start = time.perf_counter()
+            operator = operator_cls(step.params)
             image = operator.compute(image)
             t_step_end = time.perf_counter()
-            step_timings.append(StepTiming(step=i + 1, type=step.type, duration_ms=(t_step_end - t_step_start) * 1000))
+            step_timings.append(
+                StepTiming(step=i + 1, operator_type=step.type, duration_ms=(t_step_end - t_step_start) * 1000)
+            )
         except Exception as e:
             t_fail = time.perf_counter()
             return PipelineResponse(
                 success=False,
                 error=f"Error in step {i + 1} ({step.type}): {type(e).__name__}: {e}",
-                step=i,
+                step=i + 1,
                 timings=PipelineTimings(total_ms=(t_fail - t_start_total) * 1000, steps=step_timings),
             )
 
--- a/imagelab-frontend/src/components/Preview/PreviewPane.tsx
+++ b/imagelab-frontend/src/components/Preview/PreviewPane.tsx
@@ -1,5 +1,5 @@
 import { useState } from "react";
-import { ZoomIn, ZoomOut, Image, ImageDown, Trash2 } from "lucide-react";
+import { ZoomIn, ZoomOut, Image, ImageDown, Trash2, Timer } from "lucide-react";
 import { usePipelineStore } from "../../store/pipelineStore";
 import ImageDisplay from "./ImageDisplay";
 
@@ -32,6 +32,12 @@ function ZoomControls({
       </button>
     </div>
   );
+}
+
+// Operator types follow 'category_operationName' convention; strip the category prefix for display.
+function getStepLabel(operatorType: string): string {
+  const underscoreIndex = operatorType.indexOf("_");
+  return underscoreIndex !== -1 ? operatorType.slice(underscoreIndex + 1) : operatorType;
 }
 
 export default function PreviewPane() {
@@ -84,9 +90,16 @@ export default function PreviewPane() {
             <h2 className="text-xs font-semibold text-gray-500 uppercase tracking-wider">
               Processed
             </h2>
-            {timings && (
+            {timings && !error && (
               <span className="inline-flex items-center gap-1 px-2 py-0.5 rounded-full bg-green-50 border border-green-200 text-[11px] font-medium text-green-700 ml-1 mt-[-1px]">
-                ⏱ {timings.total_ms.toFixed(1)} ms
+                <Timer size={10} className="text-green-600" />
+                {timings.total_ms.toFixed(1)} ms
+              </span>
+            )}
+            {timings && error && (
+              <span className="inline-flex items-center gap-1 px-2 py-0.5 rounded-full bg-amber-50 border border-amber-200 text-[11px] font-medium text-amber-700 ml-1 mt-[-1px]">
+                <Timer size={10} className="text-amber-600" />
+                {timings.total_ms.toFixed(1)} ms (partial)
               </span>
             )}
           </div>
@@ -115,30 +128,35 @@ export default function PreviewPane() {
                 Step Timings
               </summary>
               <div className="mt-1.5 space-y-1">
-                {timings.steps.map((t) => {
-                  const label = t.type.includes("_")
-                    ? t.type.split("_").slice(1).join("_")
-                    : t.type;
-                  const maxMs = Math.max(...timings.steps.map((s) => s.duration_ms));
-                  const barWidth = maxMs > 0 ? (t.duration_ms / maxMs) * 100 : 0;
-                  return (
-                    <div key={t.step} className="flex items-center gap-2 text-[11px] text-gray-500">
-                      <span className="w-4 text-right text-gray-400 flex-shrink-0">{t.step}.</span>
-                      <span className="truncate flex-1 pr-1" title={t.type}>
-                        {label}
-                      </span>
-                      <div className="w-16 h-1.5 bg-gray-100 rounded-full flex-shrink-0">
-                        <div
-                          className="h-full bg-indigo-400 rounded-full"
-                          style={{ width: `${barWidth}%` }}
-                        />
+                {(() => {
+                  const maxMs = timings.steps.reduce((max, s) => Math.max(max, s.duration_ms), 0);
+                  return timings.steps.map((t) => {
+                    const label = getStepLabel(t.operator_type);
+                    const barWidth = maxMs > 0 ? (t.duration_ms / maxMs) * 100 : 0;
+                    return (
+                      <div
+                        key={t.step}
+                        className="flex items-center gap-2 text-[11px] text-gray-500"
+                      >
+                        <span className="w-4 text-right text-gray-400 flex-shrink-0">
+                          {t.step}.
+                        </span>
+                        <span className="truncate flex-1 pr-1" title={t.operator_type}>
+                          {label}
+                        </span>
+                        <div className="w-16 h-1.5 bg-gray-100 rounded-full flex-shrink-0">
+                          <div
+                            className="h-full bg-indigo-400 rounded-full"
+                            style={{ width: `${barWidth}%` }}
+                          />
+                        </div>
+                        <span className="flex-shrink-0 text-gray-600 w-14 text-right">
+                          {t.duration_ms.toFixed(1)} ms
+                        </span>
                       </div>
-                      <span className="flex-shrink-0 text-gray-600 w-14 text-right">
-                        {t.duration_ms.toFixed(1)} ms
-                      </span>
-                    </div>
-                  );
-                })}
+                    );
+                  });
+                })()}
               </div>
             </details>
           </div>
--- a/imagelab-backend/tests/test_pipeline_timing.py
+++ b/imagelab-backend/tests/test_pipeline_timing.py
@@ -1,0 +1,55 @@
+EXECUTE_URL = "/api/pipeline/execute"
+
+GRAY_STEP = {"type": "imageconvertions_grayimage", "params": {}}
+BINARY_STEP = {"type": "imageconvertions_graytobinary", "params": {"thresholdValue": 127, "maxValue": 255}}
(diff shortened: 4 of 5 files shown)


imagelab-frontend/src/components/Preview/PreviewPane.tsx
```diff
@@ -78,11 +78,18 @@ export default function PreviewPane() {
...
-            Processed
-          </h2>
+        <div className="flex items-center justify-between px-3 py-1.5 border-b border-gray-200">
+          <div className="flex items-center gap-1.5">
+            <ImageDown size={14} className="text-gray-400" />
+            <h2 className="text-xs font-semibold text-gray-500 uppercase tracking-wider">
+              Processed
+            </h2>
+            {timings && (
+              <span className="inline-flex items-center gap-1 px-2 py-0.5 rounded-full bg-green-50 border border-green-200 text-[11px] font-medium text-green-700 ml-1 mt-[-1px]">
```
ivantha: **[MEDIUM]** Success-styled timing badge renders during error/partial-timing states

The timing badge in the "Processed" header uses an unambiguously success-oriented green style:

```tsx
{timings && (
  <span className="... bg-green-50 border border-green-200 text-green-700 ...">
    ⏱ {timings.total_ms.toFixed(1)} ms
  </span>
)}
```

However, `timings` is non-null even when the pipeline **failed** — the PR explicitly surfaces partial timings on error. This means a user will see a green "⏱ 42.0 ms" badge alongside a red error message, which is misleading: the green badge implies a successful result, but the image was never produced.

Furthermore, the badge sits inside the **"Processed"** section header, but on failure there is no processed image in that section.

**Suggestion:**
```
Gate the success-styled badge on the absence of an error, or use a neutral/warning style for partial timings:

```tsx
{timings && !error && (
  <span className="inline-flex items-center gap-1 px-2 py-0.5 rounded-full bg-green-50 border border-green-200 text-[11px] font-medium text-green-700 ml-1">
    ⏱ {timings.total_ms.toFixed(1)} ms
  </span>
)}
{timings && error && (
  <span className="inline-flex items-center gap-1 px-2 py-0.5 rounded-full bg-amber-50 border border-amber-200 text-[11px] font-medium text-amber-700 ml-1">
    ⏱ {timings.total_ms.toFixed(1)} ms (partial)
  </span>
)}
```
```

File: imagelab-frontend/src/components/Preview/PreviewPane.tsx [hunk 2]
```diff
@@ -101,6 +108,41 @@ export default function PreviewPane() {
             <p className="text-xs text-red-600">{error}</p>
           </div>
         )}
+        {timings && timings.steps.length > 0 && (
+          <div className="px-3 py-2 bg-white border-t border-gray-200">
+            <details className="group">
+              <summary className="text-[10px] uppercase font-semibold text-gray-500 hover:text-indigo-600 cursor-pointer select-none">
+                Step Timings
+              </summary>
```
ivantha: **[MEDIUM]** Implicit, undocumented label-stripping convention creates fragile display logic

The step label is derived by stripping the first underscore-delimited segment from `t.type`:

```tsx
const label = t.type.includes("_")
  ? t.type.split("_").slice(1).join("_")
  : t.type;
```

This silently encodes an undocumented convention: operator type names are expected to follow a `category_name` format (e.g. `geometric_resize` → displays as `resize`). Issues:
1. The convention is not captured anywhere in the type system or backend models.
2. Types without underscores show the full name (`blur` → `blur`) while prefixed types show without prefix (`geometric_resize` → `resize`), so visual style is inconsistent if mixed types are present.
3. Multi-word suffixes like `geometric_to_grayscale` become `to_grayscale` — correct here, but purely by accident.
4. If the backend naming convention changes, labels will silently break.

**Suggestion:**
```
The most robust fix is to add an explicit `label` or `display_name` field to `StepTiming` in the backend:

```python
class StepTiming(BaseModel):
    step: int
    type: str
    label: str  # Human-readable display name
    duration_ms: float
```

Alternatively, at minimum add a comment explaining the convention and move the logic into a named utility function:

```tsx
// Operator types follow 'category_operationName' convention; strip the category prefix.
function getStepLabel(operatorType: string): string {
  const underscoreIndex = operatorType.indexOf("_");
  return underscoreIndex !== -1
    ? operatorType.slice(underscoreIndex + 1)
    : operatorType;
}
```
```

File: imagelab-backend/app/services/pipeline_executor.py [hunk 3]
```diff
@@ -1,46 +1,73 @@
...
+            timings=PipelineTimings(total_ms=(t_fail - t_start_total) * 1000, steps=step_timings),
+        )
+
+    t_end_total = time.perf_counter()
 
     return PipelineResponse(
         success=True,
         image=encoded,
         image_format=request.image_format,
+        timings=PipelineTimings(total_ms=(t_end_total - t_start_total) * 1000, steps=step_timings),
```
ivantha: **[MEDIUM]** No automated tests added for the new timing logic

The PR checklist explicitly marks "New tests added" as unchecked. The timing logic has several critical behaviors that are not exercised by any existing test:

1. **Partial timings on failure** — the core acceptance criterion: when a step raises an exception, the already-completed `step_timings` must be included in the response.
2. **`total_ms` accuracy** — that `total_ms` reflects wall time from the first `perf_counter` call to the last.
3. **`step` numbering** — that `StepTiming.step` values are correct and sequential.
4. **Edge cases** — empty pipeline (no steps), single-step pipeline, pipeline with NOOP steps (NOOP steps are skipped from timing).

The absence of tests is particularly concerning given finding #1 (the step-indexing bug), which a simple unit test would have caught immediately.

**Suggestion:**
```
Add tests to the existing backend test suite (e.g., `tests/test_pipeline_executor.py`):

```python
def test_successful_pipeline_includes_timings():
    # build a minimal request with one real operator
    response = execute_pipeline(request)
    assert response.timings is not None
    assert response.timings.total_ms > 0
    assert len(response.timings.steps) == 1
    assert response.timings.steps[0].step == 1

def test_partial_timings_returned_on_failure():
    # pipeline with one good step, one bad step
    response = execute_pipeline(request)
    assert response.success is False
    assert response.timings is not None
    assert len(response.timings.steps) == 1  # only the completed step
    assert response.timings.steps[0].step == 1

def test_noop_steps_excluded_from_timings():
    # pipeline with only NOOP steps
    response = execute_pipeline(request)
    assert response.timings.steps == []
```
```

File: imagelab-backend/app/services/pipeline_executor.py [hunk 4]
```diff
@@ -1,46 +1,73 @@
...
             )
 
         try:
             operator = operator_cls(step.params)
+            t_step_start = time.perf_counter()
             image = operator.compute(image)
+            t_step_end = time.perf_counter()
+            step_timings.append(StepTiming(step=i + 1, type=step.type, duration_ms=(t_step_end - t_step_start) * 1000))
         except Exception as e:
+            t_fail = time.perf_counter()
```
ivantha: **[LOW]** Operator instantiation time is excluded from per-step timing

The timing clock starts *after* `operator_cls(step.params)` completes:

```python
operator = operator_cls(step.params)   # ← not timed
t_step_start = time.perf_counter()
image = operator.compute(image)        # ← timed
t_step_end = time.perf_counter()
```

For operators that perform expensive initialization (loading weights, allocating GPU memory, reading config files, etc.), the reported `duration_ms` will significantly under-count the true cost. Users will see a fast-looking step that actually consumed far more wall time.

Even for lightweight operators, the construction cost is conceptually part of "running step N" and should be accounted for.

**Suggestion:**
```
Move `t_step_start` before operator instantiation:

```python
try:
    t_step_start = time.perf_counter()       # ← start before construction
    operator = operator_cls(step.params)
    image = operator.compute(image)
    t_step_end = time.perf_counter()
    step_timings.append(
        StepTiming(step=i + 1, type=step.type,
                   duration_ms=(t_step_end - t_step_start) * 1000)
    )
except Exception as e:
    ...
```

If there's a deliberate reason to separate construction from computation, the model should expose both fields (`init_ms` and `compute_ms`) so users can distinguish between them.
```

File: imagelab-backend/app/models/pipeline.py [hunk 5]
```diff
@@ -12,9 +12,21 @@ class PipelineRequest(BaseModel):
     pipeline: list[PipelineStep]
 
 
+class StepTiming(BaseModel):
+    step: int
```
ivantha: **[LOW]** `type` attribute in `StepTiming` shadows Python's built-in

The field name `type: str` in `StepTiming` shadows the Python built-in `type()` function at the instance attribute level. While Pydantic v2 handles this correctly and it is technically valid Python, it can cause surprising behaviour if someone writes `type(instance.type)` intending to check the Python type of the attribute, accidentally accessing the field value instead. It also confuses linters and some IDE completion engines.

This is a minor ergonomic concern, but naming fields `type` is a known footgun in Python codebases.

**Suggestion:**
```
Consider renaming to `operator_type` or `block_type` to be unambiguous and more descriptive:

```python
class StepTiming(BaseModel):
    step: int
    operator_type: str   # was: type
    duration_ms: float
```

Update the frontend `StepTiming` interface accordingly:
```typescript
export interface StepTiming {
  step: number;
  operator_type: string;  // was: type
  duration_ms: number;
}
```
```

File: imagelab-frontend/src/components/Preview/PreviewPane.tsx [hunk 6]
```diff
@@ -78,11 +78,18 @@ export default function PreviewPane() {
...
-          <ImageDown size={14} className="text-gray-400" />
-          <h2 className="text-xs font-semibold text-gray-500 uppercase tracking-wider">
-            Processed
-          </h2>
+        <div className="flex items-center justify-between px-3 py-1.5 border-b border-gray-200">
+          <div className="flex items-center gap-1.5">
+            <ImageDown size={14} className="text-gray-400" />
+            <h2 className="text-xs font-semibold text-gray-500 uppercase tracking-wider">
+              Processed
+            </h2>
```
ivantha: **[INFO]** Raw Unicode emoji used instead of a Lucide icon component

The timing badge uses a raw Unicode emoji character `⏱` directly in JSX:

```tsx
⏱ {timings.total_ms.toFixed(1)} ms
```

The rest of the codebase already imports icon components from `lucide-react` (e.g., `ImageDown`, `RefreshCw` visible in the same file). Using a raw emoji introduces a visual inconsistency — emoji rendering varies across operating systems, browsers, and fonts, whereas Lucide icons are rendered as SVG and are pixel-consistent.

**Suggestion:**
```
Replace the emoji with an equivalent Lucide icon, for example `Timer` or `Clock`:

```tsx
import { ImageDown, Timer } from "lucide-react";

// In the badge:
<span className="inline-flex items-center gap-1 ...">
  <Timer size={10} className="text-green-600" />
  {timings.total_ms.toFixed(1)} ms
</span>
```
```

File: imagelab-frontend/src/components/Preview/PreviewPane.tsx [hunk 7]
```diff
@@ -101,6 +108,41 @@ export default function PreviewPane() {
...
+                    </div>
+                  );
+                })}
+              </div>
+            </details>
+          </div>
+        )}
         <ZoomControls
           disabled={!processedImage}
           onZoomIn={zoomIn(setProcessedZoom)}
```
ivantha: **[INFO]** Positive: Partial timings on failure is an excellent design decision

Returning and surfacing partial `step_timings` even when the pipeline errors out is a genuinely thoughtful design choice. The backend accumulates completed step timings in a list and includes them in every error response, giving users actionable insight into exactly how far execution reached and where time was spent before failure. This directly fulfills the acceptance criterion without any special-casing on the frontend — the same rendering code handles both complete and partial timing lists. The approach is clean, maintainable, and user-friendly.

**Suggestion:**
```
No change needed. Consider noting the partial-timing contract explicitly in the docstring of `execute_pipeline` so future maintainers understand the invariant:

```python
def execute_pipeline(request: PipelineRequest) -> PipelineResponse:
    """
    Execute the image-processing pipeline described by *request*.

    Returns a PipelineResponse that always includes a ``timings`` field
    populated with every step that completed before the function returned,
    even when the response indicates failure.  This allows callers to
    inspect partial execution progress on error.
    """
```
```

File: imagelab-frontend/src/components/Layout.tsx [hunk 8]
```diff
@@ -9,10 +10,12 @@ import { ErrorBoundary } from "./ErrorBoundary";
 
 export default function Layout() {
   const { containerRef, workspace } = useBlocklyWorkspace();
+  const { reset } = usePipelineStore();
   const [resetKey, setResetKey] = useState(0);
 
   const handleEditorReset = () => {
```
ivantha: **[INFO]** Positive: `handleEditorReset` now correctly clears pipeline store state

The addition of `reset()` to `handleEditorReset` ensures that when the editor is forcibly re-mounted (via the `resetKey` increment in the ErrorBoundary recovery path), the Zustand pipeline store — including images, errors, and the newly added timings — is also cleared. Without this, stale pipeline state (including the partial timing data this PR introduces) would persist across an error boundary reset. The ErrorBoundary button label change from "Reload Application" to "Reset Editor" accurately describes the narrower scope of the action, avoiding a potentially surprising full-page reload. This is a coherent and well-considered improvement.

**Suggestion:**
```
No change needed. Worth double-checking that `ErrorBoundary` passes `onReset` to the button's `onClick` handler (not visible in the diff but implied by the test setup) so `handleEditorReset` is actually invoked, and not just `window.location.reload()`.
```

PR #81, opened 2026-02-27
add input validation to blur operators

### LINKED ISSUE #44: Add Input Validation to Blur Operators
**Difficulty:** Medium
**Area:** backend

## Description

The three blur operators (`blur.py`, `gaussian_blur.py`, `median_blur.py`) currently accept raw parameters from the pipeline request without any validation. This can lead to cryptic OpenCV errors when users provide invalid values such as even kernel sizes, zero dimensions, or negative numbers. OpenCV requires kernel sizes to be positive odd integers for Gaussian and median blur operations, and general blur requires positive dimensions.

Adding proper input validation with clear, user-friendly error messages will improve the developer experience and make debugging pipelines much easier. Validated operators should raise `ValueError` with descriptive messages that explain what was wrong and what values are acceptable.

## Acceptance Criteria

- [ ] All three blur operators validate their parameters before processing and reject invalid values with clear, user-friendly error messages
- [ ] Kernel sizes are validated as positive integers; Gaussian blur and median blur additionally require odd kernel sizes
- [ ] Validation logic is shared across operators to avoid code duplication
- [ ] Default parameter values remain unchanged and valid
- [ ] Existing valid pipelines continue to work without any changes

## Resources

- [OpenCV GaussianBlur requirements](https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#gaabe8c836e97159a9193fb0b11ac52cf1) (ksize must be odd)
- [OpenCV medianBlur requirements](https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#ga564869aa33e58769b4013e36f7b7de89) (ksize must be odd and greater than 1)
- Current operators: `imagelab-backend/app/operators/blurring/`

areebahmeddd:
## Description

Adds explicit input validation to all three blur operators (`Blur`, `GaussianBlur`, `MedianBlur`) so that invalid parameters fail fast with clear, user-friendly error messages instead of producing cryptic OpenCV errors.

Validation logic is centralised in a new `validation.py` module inside the `blurring` package to avoid duplication across operators. The previous silent auto-correction behaviour in `GaussianBlur` and `MedianBlur` (silently incrementing even kernel sizes) has been replaced with explicit errors that tell the user exactly what went wrong and what valid values look like.

Fixes #44

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

91 unit tests added across four new test files:

- `tests/operators/blurring/test_validation.py` — isolates each validator function
- `tests/operators/blurring/test_blur.py`
- `tests/operators/blurring/test_gaussian_blur.py`
- `tests/operators/blurring/test_median_blur.py`

Each file covers valid inputs, every invalid category (zero, negative, even where odd is required), error message content, and explicit regression tests confirming the old silent auto-fix no longer occurs.

- [ ] Existing tests pass
- [x] New tests added
- [ ] Manual testing

N/A

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/blurring/validation.py
+++ b/imagelab-backend/app/operators/blurring/validation.py
@@ -1,25 +1,77 @@
+"""
+Input validators for blurring operator parameters.
+
+Each function raises ValueError with a user-friendly message if the provided
+value violates the constraint required by the underlying OpenCV call. Callers
+are expected to cast raw parameter values to int before calling these helpers.
+
+Validators:
+  validate_positive_kernel_dim      - used by Blur (widthSize, heightSize)
+  validate_positive_odd_kernel_size - used by GaussianBlur (widthSize, heightSize)
+  validate_median_kernel_size       - used by MedianBlur (kernelSize)
+"""
+
+
+def _require_int(value: object, name: str) -> None:
+    """Raise TypeError if value is not a plain int.
+
+    Python does not enforce type annotations at runtime, so float inputs like
+    3.0 or 2.5 would otherwise slip through the odd/positive checks. This
+    guard must be called before any arithmetic validation.
+    """
+    if not isinstance(value, int):
+        raise TypeError(f"'{name}' must be an integer, got {type(value).__name__}: {value!r}")
+
+
+def _validate_odd_kernel(value: int, name: str, min_value: int, examples: str) -> None:
+    """Shared logic for odd-kernel validators.
+
+    Raises ValueError when value is below min_value or even. The lower-neighbour
+    suggestion for even values is only included when it is itself >= min_value,
+    preventing suggestions of invalid values (e.g. when value=2 and min_value=3,
+    only 3 is suggested, never 1).
+    """
+    if value < min_value:
+        raise ValueError(
+            f"'{name}' must be a positive odd integer >= {min_value}, got {value}. Use a value like {examples}."
+        )
+    if value % 2 == 0:
+        lower = value - 1
+        upper = value + 1
+        suggestion = f"{lower} or {upper}" if lower >= min_value else str(upper)
+        raise ValueError(
+            f"'{name}' must be a positive odd integer >= {min_value}, got {value} (even). Did you mean {suggestion}?"
+        )
+
+
 def validate_positive_kernel_dim(value: int, name: str) -> None:
-    """Raise ValueError if value is not a positive integer."""
+    """Raise ValueError if value is not a positive integer.
+
+    Used by Blur for widthSize and heightSize.
+    """
+    _require_int(value, name)
     if value <= 0:
         raise ValueError(f"'{name}' must be a positive integer, got {value}. Use a value of 1 or greater.")
 
 
 def validate_positive_odd_kernel_size(value: int, name: str) -> None:
-    """Raise ValueError if value is not a positive odd integer (required by GaussianBlur)."""
-    if value <= 0:
-        raise ValueError(f"'{name}' must be a positive odd integer, got {value}. Use a value like 1, 3, 5, 7, …")
-    if value % 2 == 0:
-        raise ValueError(
-            f"'{name}' must be a positive odd integer, got {value} (even). Did you mean {value - 1} or {value + 1}?"
-        )
+    """Raise ValueError if value is not a positive odd integer.
+
+    Used by GaussianBlur for widthSize and heightSize. OpenCV requires the
+    kernel size to be a positive odd integer.
+    """
+    _require_int(value, name)
+    _validate_odd_kernel(value, name, min_value=1, examples="1, 3, 5, 7")
 
 
 def validate_median_kernel_size(value: int, name: str) -> None:
-    """Raise ValueError if value is not an odd integer greater than 1 (required by medianBlur)."""
-    if value <= 1:
-        raise ValueError(f"'{name}' must be an odd integer greater than 1, got {value}. Use a value like 3, 5, 7, …")
-    if value % 2 == 0:
-        raise ValueError(
-            f"'{name}' must be an odd integer greater than 1, got {value} (even). "
-            f"Did you mean {value - 1} or {value + 1}?"
-        )
+    """Raise ValueError if value is not a valid MedianBlur kernel size.
+
+    OpenCV technically accepts ksize=1 (identity operation), but this validator
+    rejects it to prevent accidentally passing a no-op blur. Valid values are
+    odd integers >= 3 (e.g. 3, 5, 7).
+
+    Used by MedianBlur for kernelSize.
+    """
+    _require_int(value, name)
+    _validate_odd_kernel(value, name, min_value=3, examples="3, 5, 7")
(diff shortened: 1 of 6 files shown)


imagelab-backend/app/operators/blurring/validation.py
```diff
@@ -0,0 +1,25 @@
...
+
+def validate_median_kernel_size(value: int, name: str) -> None:
+    """Raise ValueError if value is not an odd integer greater than 1 (required by medianBlur)."""
+    if value <= 1:
+        raise ValueError(f"'{name}' must be an odd integer greater than 1, got {value}. Use a value like 3, 5, 7, …")
+    if value % 2 == 0:
+        raise ValueError(
+            f"'{name}' must be an odd integer greater than 1, got {value} (even). "
+            f"Did you mean {value - 1} or {value + 1}?"
+        )
```
ivantha: **[LOW]** No runtime type enforcement — float inputs bypass the odd/positive checks

All three validator functions are annotated `value: int` but Python does not enforce type annotations at runtime. A caller passing a float slips through undetected:

```python
validate_positive_odd_kernel_size(3.0, "widthSize")  # passes — 3.0 % 2 == 1.0 != 0
validate_positive_odd_kernel_size(2.5, "widthSize")  # passes — 2.5 % 2 == 0.5 != 0
```

This matters because `self.params.get(...)` returns whatever the caller serialised (could be a JSON float), and the `int(...)` cast in the operators truncates rather than rejects (e.g., `int(2.9)` → `2`). While the operators do cast before calling, the validators are in a public module and could be called directly by future code without the cast.

**Suggestion:**
```
Add an `isinstance` guard at the top of each validator, or create a shared private helper:

```python
def _require_int(value: object, name: str) -> int:
    if not isinstance(value, int):
        raise TypeError(
            f"'{name}' must be an integer, got {type(value).__name__}: {value!r}"
        )
    return value

def validate_positive_kernel_dim(value: int, name: str) -> None:
    _require_int(value, name)
    if value <= 0:
        ...
```
```

File: imagelab-backend/app/operators/blurring/validation.py [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+    if value % 2 == 0:
+        raise ValueError(
+            f"'{name}' must be a positive odd integer, got {value} (even). Did you mean {value - 1} or {value + 1}?"
+        )
+
+
+def validate_median_kernel_size(value: int, name: str) -> None:
+    """Raise ValueError if value is not an odd integer greater than 1 (required by medianBlur)."""
+    if value <= 1:
+        raise ValueError(f"'{name}' must be an odd integer greater than 1, got {value}. Use a value like 3, 5, 7, …")
```
ivantha: **[LOW]** `validate_median_kernel_size` is more restrictive than OpenCV — undocumented design decision

`cv2.medianBlur` with `ksize = 1` is valid in OpenCV; it simply returns the original image unchanged (a no-op blur). The validator rejects `ksize = 1` with the constraint "must be an odd integer **greater than 1**". This is a deliberate design choice (probably to prevent user confusion), but:

1. The docstring only says `"required by medianBlur"`, which is factually inaccurate — OpenCV itself allows it.
2. There is no inline comment explaining *why* 1 is excluded.
3. It introduces an asymmetry: `GaussianBlur` accepts `ksize = 1` (valid and treated as a no-op) but `MedianBlur` does not.

**Suggestion:**
```
Either:
- **Allow 1**: change the lower bound to `<= 0` to match the actual OpenCV constraint, or
- **Keep the restriction**: update the docstring to accurately reflect the design rationale:

```python
def validate_median_kernel_size(value: int, name: str) -> None:
    """
    Raise ValueError if value is not a valid MedianBlur kernel size.

    OpenCV technically accepts ksize=1 (identity operation), but this
    validator rejects it to prevent accidentally passing a no-op blur.
    Valid values are odd integers >= 3 (e.g. 3, 5, 7).
    """
```
```

File: imagelab-backend/tests/operators/blurring/test_blur.py [hunk 3]
```diff
@@ -0,0 +1,43 @@
+import numpy as np
+import pytest
+
+from app.operators.blurring.blur import Blur
+
+# A tiny valid image used across all tests
```
ivantha: **[LOW]** Module-level `IMAGE` constant is shared across all test methods — consider using a fixture

All four test files define a module-level `IMAGE = np.full(...)` constant. While `cv2.blur` / `cv2.GaussianBlur` / `cv2.medianBlur` do not mutate their input arrays in-place (they return new arrays), making the current code safe, relying on this behaviour is an implicit assumption:

1. If a future test uses an in-place operation (e.g., `cv2.blur(IMAGE, ..., dst=IMAGE)`), all subsequent tests in the module will see the mutated image.
2. The constant is defined at import time, so if the fixture needs parametrization later (e.g., grayscale vs colour, different dtypes), the module constant becomes an obstacle.
3. pytest fixtures with appropriate scope (`function` or `module`) make the intent explicit and the lifetime obvious.

**Suggestion:**
```
Replace the module-level constant with a function-scoped fixture:

```python
import pytest
import numpy as np

@pytest.fixture
def image():
    return np.full((5, 5, 3), 128, dtype=np.uint8)

class TestBlurValidInput:
    def test_default_params_produce_output(self, image):
        result = Blur({}).compute(image)
        assert result.shape == image.shape
```
```

File: imagelab-backend/tests/operators/blurring/test_gaussian_blur.py [hunk 4]
```diff
@@ -0,0 +1,58 @@
...
+        with pytest.raises(ValueError, match="3|5"):
+            GaussianBlur({"widthSize": 4, "heightSize": 3}).compute(IMAGE)
+
+    def test_error_message_is_user_friendly(self):
+        with pytest.raises(ValueError, match="odd"):
+            GaussianBlur({"widthSize": 2}).compute(IMAGE)
+
+    def test_no_silent_autofix_for_even_input(self):
+        # Previously, even values were silently incremented to odd. Now a
+        # ValueError must be raised instead of silently accepting the input.
```
ivantha: **[LOW]** Test for `test_even_error_suggests_neighbours` only covers `value = 4`, not `value = 2`

The test `test_even_error_suggests_neighbours` asserts that the error message for `widthSize = 4` contains `3` or `5`. This is correct, but it only validates one specific mid-range even value. It does not verify:
- `value = 2` (boundary: suggests 1 and 3 for GaussianBlur — both valid).
- Very large even values (e.g., 100 → suggests 99 or 101).

For GaussianBlur's validator this is not a bug (1 and 3 are both valid), but the lack of coverage means a future regression could go unnoticed.

**Suggestion:**
```
Parametrize the neighbour test:

```python
@pytest.mark.parametrize("even, expected_lower, expected_upper", [
    (2, 1, 3),
    (4, 3, 5),
    (100, 99, 101),
])
def test_even_error_suggests_neighbours(self, even, expected_lower, expected_upper):
    with pytest.raises(ValueError, match=f"{expected_lower}|{expected_upper}"):
        GaussianBlur({"widthSize": even, "heightSize": 3}).compute(IMAGE)
```
```

imagelab-backend/app/operators/blurring/validation.py
areebahmeddd: `_require_int` handles that

PR #83, opened 2026-02-27
Add Image Resize Operator (Absolute Dimensions) #53

### LINKED ISSUE #53: Add Image Resize Operator (Absolute Dimensions)
**Difficulty:** Medium
**Area:** backend, frontend

## Description

ImageLab currently has a scale operator that resizes images using multiplication factors (e.g., 2x, 0.5x). However, users often need to resize images to specific pixel dimensions (e.g., 640x480, 1920x1080). This is common when preparing images for machine learning pipelines, web display, or matching specific resolution requirements.

The resize operator should accept absolute width and height values in pixels and apply `cv2.resize()`. It should also expose an interpolation method parameter, since different interpolation algorithms are optimal for different scenarios: `INTER_LINEAR` for general use, `INTER_AREA` for shrinking, and `INTER_CUBIC` or `INTER_LANCZOS4` for enlarging.

## Acceptance Criteria

- [ ] A resize operator is implemented in the backend following the existing operator pattern
- [ ] The operator accepts width, height, and interpolation parameters, with dimensions in absolute pixels (not scale factors)
- [ ] Five interpolation methods are supported: LINEAR, AREA, CUBIC, NEAREST, LANCZOS4
- [ ] Width and height are validated to be at least 1; defaults produce a no-op (current image size)
- [ ] The operator is registered and available in the processing pipeline
- [ ] A Blockly block exists with numeric fields for dimensions and an interpolation dropdown, appearing in the geometric category
- [ ] This operator is clearly distinct from the existing scale operator in naming and presentation

## Resources

- [cv2.resize API reference](https://docs.opencv.org/4.x/da/d54/group__imgproc__transform.html#ga47a974309e9102f5f08231edc7e7529d)
- [OpenCV interpolation flags](https://docs.opencv.org/4.x/da/d54/group__imgproc__transform.html#ga5bb5a1fea74ea38e1a5445f6aad53fdb)
- Existing scale operator: `imagelab-backend/app/operators/geometric/scale_image.py`

Jayant-kernel:
## Description

Added the `geometric_resizeimage` operator to explicitly resize images to precise absolute dimensions (width/height in real pixels) using `cv2.resize()`, addressing the limitations of the existing generalized scalar multiplier.
- Validates dimensions and safely performs no-ops if boundaries are missing or negative natively on the backend.
- Exposes all 5 standard OpenCV interpolations: `LINEAR`, `AREA`, `CUBIC`, `NEAREST`, `LANCZOS4`.
- Frontend registers a distinct "Resize Image (Pixels)" UI block natively injected within the Geometric drawer, fully equipped with dropdowns and correct property typings natively identical to its backend counterpart registry key. 

Fixes #53

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

The `imagelab-frontend` definitions were verified fully using Node + Vite without typescript degradation (`npm run build`). The backend core has unit-tests implemented simulating single boundary drops, negative bounds safely recovering to origin bounds iteratively passing the pipeline.

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/geometric/resize_image.py
+++ b/imagelab-backend/app/operators/geometric/resize_image.py
@@ -3,6 +3,14 @@ import numpy as np
 
 from app.operators.base import BaseOperator
 
+_INTERPOLATION_MAP: dict[str, int] = {
+    "LINEAR": cv2.INTER_LINEAR,
+    "AREA": cv2.INTER_AREA,
+    "CUBIC": cv2.INTER_CUBIC,
+    "NEAREST": cv2.INTER_NEAREST,
+    "LANCZOS4": cv2.INTER_LANCZOS4,
+}
+
 
 class ResizeImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
@@ -10,8 +18,8 @@ class ResizeImage(BaseOperator):
         original_rows, original_cols = image.shape[:2]
 
         try:
-            width = int(self.params.get("width", original_cols))
-            height = int(self.params.get("height", original_rows))
+            width = int(round(float(self.params.get("width", original_cols))))
+            height = int(round(float(self.params.get("height", original_rows))))
         except (ValueError, TypeError):
             width = original_cols
             height = original_rows
@@ -22,21 +30,13 @@ class ResizeImage(BaseOperator):
         if height < 1:
             height = original_rows
 
-        # Pure No-op shortcut
-        if width == original_cols and height == original_rows:
-            return image
-
+        # Resolve interpolation before the no-op check so invalid strings are never silently ignored
         interpolation_method_str = str(self.params.get("interpolation", "LINEAR")).upper()
-
-        interpolation_map = {
-            "LINEAR": cv2.INTER_LINEAR,
-            "AREA": cv2.INTER_AREA,
-            "CUBIC": cv2.INTER_CUBIC,
-            "NEAREST": cv2.INTER_NEAREST,
-            "LANCZOS4": cv2.INTER_LANCZOS4,
-        }
-
         # Default to INTER_LINEAR if the method isn't explicitly found
-        interpolation_flag = interpolation_map.get(interpolation_method_str, cv2.INTER_LINEAR)
+        interpolation_flag = _INTERPOLATION_MAP.get(interpolation_method_str, cv2.INTER_LINEAR)
+
+        # Pure No-op shortcut — return a copy for consistent ownership semantics
+        if width == original_cols and height == original_rows:
+            return image.copy()
 
         return cv2.resize(image, (width, height), interpolation=interpolation_flag)
--- a/imagelab-backend/tests/test_resize_image.py
+++ b/imagelab-backend/tests/test_resize_image.py
@@ -1,5 +1,4 @@
 import numpy as np
-import pytest
 
 from app.operators.geometric.resize_image import ResizeImage
 
@@ -37,8 +36,41 @@ def test_resize_interpolation_methods():
 
     for method in ["LINEAR", "AREA", "CUBIC", "NEAREST", "LANCZOS4"]:
         op = ResizeImage(params={"width": 15, "height": 15, "interpolation": method})
-        try:
-            result = op.compute(image)
-            assert result.shape == (15, 15, 3), f"{method} interpolation produced incorrect shape"
-        except Exception as e:
-            pytest.fail(f"Interpolation {method} threw an exception: {e}")
+        result = op.compute(image)
+        assert result.shape == (15, 15, 3), f"{method} interpolation produced incorrect shape"
+
+
+def test_resize_zero_dimension_fallback():
+    image = np.zeros((15, 25, 3), dtype=np.uint8)
+
+    op = ResizeImage(params={"width": 0, "height": 0})
+
+    result = op.compute(image)
+    assert result.shape == (15, 25, 3), "Zero dimensions should fall back to original"
+
+
+def test_resize_partial_zero_dimension():
+    image = np.zeros((15, 25, 3), dtype=np.uint8)
+
+    op = ResizeImage(params={"width": 0, "height": 30})
+
+    result = op.compute(image)
+    assert result.shape == (30, 25, 3), "Zero width falls back; valid height applies"
+
+
+def test_resize_unknown_interpolation_defaults_to_linear():
+    image = np.zeros((10, 10, 3), dtype=np.uint8)
+
+    op = ResizeImage(params={"width": 20, "height": 20, "interpolation": "BILINEAR"})
+
+    result = op.compute(image)  # Should not raise; falls back to LINEAR
+    assert result.shape == (20, 20, 3)
+
+
+def test_resize_grayscale_image():
+    image = np.zeros((10, 10), dtype=np.uint8)  # 2D grayscale
+
+    op = ResizeImage(params={"width": 20, "height": 30})
+
+    result = op.compute(image)
+    assert result.shape == (30, 20), "Grayscale image should resize correctly"
--- a/imagelab-frontend/src/blocks/categories.ts
+++ b/imagelab-frontend/src/blocks/categories.ts
@@ -30,7 +30,7 @@ export const categories: CategoryInfo[] = [
       { type: "geometric_rotateimage", label: "Rotate Image" },
       { type: "geometric_affineimage", label: "Affine Transform" },
       { type: "geometric_scaleimage", label: "Scale Image" },
-      { type: "geometric_resizeimage", label: "Resize Image" },
+      { type: "geometric_resizeimage", label: "Resize Image (Pixels)" },
     ],
   },
   {
--- a/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
@@ -49,8 +49,8 @@ export const geometricBlocks = [
     type: "geometric_resizeimage",
     message0: "Resize Image to width %1 and height %2 | Interpolation %3",
     args0: [
-      { type: "field_number", name: "width", value: 1, min: 0 },
-      { type: "field_number", name: "height", value: 1, min: 0 },
+      { type: "field_number", name: "width", value: 640, min: 1 },
+      { type: "field_number", name: "height", value: 480, min: 1 },
       {
         type: "field_dropdown",
         name: "interpolation",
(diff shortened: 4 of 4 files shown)


ivantha:
## ◎ Argus Review

This PR adds a well-structured `ResizeImage` operator that cleanly follows existing codebase conventions. The backend implementation is solid — parameter validation, interpolation mapping, and the no-op shortcut are all logically sound. However, there are two important frontend configuration issues: `min: 0` on the number fields silently allows zero-dimension inputs (which fall back without any user feedback), and the default `value: 1` for both width and height would reduce any untouched image to a 1×1 pixel thumbnail. The test suite covers the happy path and basic fallbacks well, but is missing several edge cases (zero dimensions, unknown interpolation strings, grayscale images). No critical bugs were found; overall quality is good with a few targeted fixes needed before merge.

### Findings (13)

#### 🟡 min: 0 allows zero-dimension input, causing silent backend fallback
**Severity:** medium | **Category:** bug | **File:** `imagelab-frontend/src/blocks/definitions/geometric.blocks.ts:52`

Both `field_number` args for `width` and `height` are configured with `min: 0`, which means a user can legally enter `0` in the UI. When `0` is sent to the backend, the condition `if width < 1` silently resets it to the original dimension — the operation behaves as a no-op **with no error or warning surfaced to the user**. This is a confusing and hard-to-diagnose failure mode: the user sets `width: 0`, expects some defined behavior, and the image appears unchanged.

The backend guard is a good last-resort safety net, but `min: 0` makes this the primary code path rather than an edge-case backstop.

**Suggestion:**
```
Change `min: 0` to `min: 1` on both fields:
```typescript
{ type: "field_number", name: "width",  value: 100, min: 1 },
{ type: "field_number", name: "height", value: 100, min: 1 },
```
```

---

#### 🟡 Default width=1, height=1 will silently destroy the image
**Severity:** medium | **Category:** bug | **File:** `imagelab-frontend/src/blocks/definitions/geometric.blocks.ts:52`

The `value: 1` default for both `width` and `height` means that if a user drags the block into the canvas and runs the pipeline without adjusting the fields, the image is resized to **1×1 pixel** — effectively destroying it. A 1-pixel image gives no visual feedback that something went wrong.

This is especially dangerous in a block-based UI where new users may not realize they need to modify every field before executing.

**Suggestion:**
```
Use a more reasonable default value unlikely to destroy data:
```typescript
{ type: "field_number", name: "width",  value: 640, min: 1 },
{ type: "field_number", name: "height", value: 480, min: 1 },
```
```

---

#### 🔵 No-op path returns a mutable reference to the original image array
**Severity:** low | **Category:** bug | **File:** `imagelab-backend/app/operators/geometric/resize_image.py:26`

`return image` at line 27 returns the **same NumPy array object** that was passed in. If any subsequent operator in the pipeline performs an in-place mutation (e.g., `image[:] = ...`), it would retroactively modify the upstream node's output. By contrast, the normal path through `cv2.resize` always returns a **new** array, creating inconsistent ownership semantics depending on whether a resize actually occurred.

**Suggestion:**
```
Return a copy to ensure consistent semantics:
```python
# Pure No-op shortcut
if width == original_cols and height == original_rows:
    return image.copy()
```
```

---

#### 🔵 Interpolation method not validated when no-op shortcut is taken
**Severity:** low | **Category:** logic | **File:** `imagelab-backend/app/operators/geometric/resize_image.py:26`

The no-op check on line 26 returns early **before** parsing or validating the `interpolation` parameter. If a user passes an invalid interpolation string (e.g. `"BILINEAR"`, `"LINAER"`) **and** the requested dimensions happen to match the current image dimensions, the invalid parameter is silently ignored. This can mask misconfigured pipelines and make debugging difficult.

**Suggestion:**
```
Resolve the interpolation flag before the no-op check:
```python
interpolation_method_str = str(self.params.get("interpolation", "LINEAR")).upper()
interpolation_flag = interpolation_map.get(interpolation_method_str, cv2.INTER_LINEAR)

# Pure No-op shortcut (placed after interpolation validation)
if width == original_cols and height == original_rows:
    return image.copy()

return cv2.resize(image, (width, height), interpolation=interpolation_flag)
```
```

---

#### 🔵 int() silently truncates float dimension values
**Severity:** low | **Category:** logic | **File:** `imagelab-backend/app/operators/geometric/resize_image.py:13`

`int(20.9)` evaluates to `20`, not `21`. If `params` receives float values for `width` or `height` (e.g., from a JavaScript frontend that sends JSON numbers without explicit integer typing), dimensions will be silently truncated. A user specifying `width: 20.9` expecting 21 pixels gets 20 with no indication the value was modified.

**Suggestion:**
```
Use explicit rounding before casting:
```python
width  = int(round(float(self.params.get("width",  original_cols))))
height = int(round(float(self.params.get("height", original_rows))))
```
If truncation is intentional, document it explicitly in a comment.
```

---

#### 🔵 No test for zero-dimension inputs (width=0, height=0)
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_resize_image.py:26`

There is a test for negative values (`-5`) but no test for `0`. The frontend's `min: 0` makes zero a legal UI input. The condition `width < 1` handles `0` and `-5` identically, but this boundary value should be explicitly tested since `0` carries its own semantic meaning ("no width") distinct from negative numbers.

**Suggestion:**
```
```python
def test_resize_zero_dimension_fallback():
    image = np.zeros((15, 25, 3), dtype=np.uint8)
    op = ResizeImage(params={"width": 0, "height": 0})
    result = op.compute(image)
    assert result.shape == (15, 25, 3), "Zero dimensions should fall back to original"

def test_resize_partial_zero_dimension():
    image = np.zeros((15, 25, 3), dtype=np.uint8)
    op = ResizeImage(params={"width": 0, "height": 30})
    result = op.compute(image)
    assert result.shape == (30, 25, 3), "Zero width falls back; valid height applies"
```
```

---

#### 🔵 No test for unknown/invalid interpolation method
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_resize_image.py:35`

There is no test verifying behavior when an unrecognized interpolation string is passed (e.g. `"BILINEAR"`, `"NONE"`, or a typo like `"LINAER"`). The current code silently defaults to `INTER_LINEAR`. Without a test, any future change to this behavior (e.g. raising a `ValueError`) would have no regression coverage, and the silent-default contract is undocumented.

**Suggestion:**
```
```python
def test_resize_unknown_interpolation_defaults_to_linear():
    image = np.zeros((10, 10, 3), dtype=np.uint8)
    op = ResizeImage(params={"width": 20, "height": 20, "interpolation": "BILINEAR"})
    result = op.compute(image)  # Should not raise; falls back to LINEAR
    assert result.shape == (20, 20, 3)
```
```

---

#### 🔵 try/except anti-pattern in test obscures failure details
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_resize_image.py:39`

Wrapping assertions inside `try/except` and calling `pytest.fail()` manually is a common anti-pattern in pytest. If `op.compute()` raises an exception, the `except` clause re-reports it without the original traceback. If the shape assertion fails, the `AssertionError` itself is caught and re-reported as a plain string, losing pytest's rich diff output.

**Suggestion:**
```
Remove the try/except and let exceptions propagate naturally:
```python
def test_resize_interpolation_methods():
    image = np.zeros((10, 10, 3), dtype=np.uint8)
    for method in ["LINEAR", "AREA", "CUBIC", "NEAREST", "LANCZOS4"]:
        op = ResizeImage(params={"width": 15, "height": 15, "interpolation": method})
        result = op.compute(image)
        assert result.shape == (15, 15, 3), f"{method} produced incorrect shape"
```
```

---

#### 🔵 No tests for grayscale (2D) or single-channel images
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_resize_image.py:1`

All four tests use 3-channel `(H, W, 3)` images. The operator calls `cv2.resize`, which works on both 2D `(H, W)` and 3D `(H, W, C)` arrays, and `image.shape[:2]` is safe for both. However, without a test, a future refactor that accidentally breaks grayscale support would go undetected.

**Suggestion:**
```
```python
def test_resize_grayscale_image():
    image = np.zeros((10, 10), dtype=np.uint8)  # 2D grayscale
    op = ResizeImage(params={"width": 20, "height": 30})
    result = op.compute(image)
    assert result.shape == (30, 20), "Grayscale image should resize correctly"
```
```

---

#### ℹ️ Positive: clean, well-structured operator implementation
**Severity:** info | **Category:** style | **File:** `imagelab-backend/app/operators/geometric/resize_image.py:1`

The operator follows the existing `BaseOperator` pattern cleanly. Parameter validation is thorough and handles missing values, invalid types, and out-of-range values **independently per axis** (a thoughtful design choice). The try/except block correctly catches both `ValueError` (e.g. `int("abc")`) and `TypeError` (e.g. `int(None)`). The registry key `geometric_resizeimage` and import ordering are consistent with adjacent entries. The no-op shortcut avoids a redundant `cv2.resize` call. Overall this is a readable, focused, well-considered implementation.

**Suggestion:**
```
No changes required — this is a positive observation recognizing good defensive programming and adherence to codebase conventions.
```

---

#### ℹ️ Tooltip uses OpenCV INTER_ prefix but dropdown options omit it — inconsistent naming
**Severity:** info | **Category:** documentation | **File:** `imagelab-frontend/src/blocks/definitions/geometric.blocks.ts:74`

The tooltip reads: *"INTER_LINEAR for general use, INTER_AREA for shrinking, INTER_CUBIC/INTER_LANCZOS4 for enlarging."* However, the dropdown options display `"LINEAR"`, `"AREA"`, `"CUBIC"` etc. — **without** the `INTER_` prefix. A user reading the tooltip and scanning the dropdown for `INTER_LINEAR` won't find it.

**Suggestion:**
```
Update the tooltip to use the same identifiers as the dropdown:
```typescript
tooltip:
  "Resizes the image to absolute dimensions in pixels. Distinct from Scale which uses a multiplier. "
  + "LINEAR for general use, AREA for shrinking, CUBIC/LANCZOS4 for high-quality enlarging.",
```
```

---

#### ℹ️ Category label 'Resize Image' does not match PR description 'Resize Image (Pixels)'
**Severity:** info | **Category:** style | **File:** `imagelab-frontend/src/blocks/categories.ts:33`

The PR description explicitly names the UI block *"Resize Image (Pixels)"* to distinguish it from *"Scale Image"* (which uses a ratio multiplier). The registered label is just `"Resize Image"`. Without the parenthetical qualifier, a user scanning the block palette side-by-side with "Scale Image" may not immediately understand the difference.

**Suggestion:**
```
Use the more descriptive label from the PR description:
```typescript
{ type: "geometric_resizeimage", label: "Resize Image (Pixels)" },
```
```

---

#### ℹ️ interpolation_map rebuilt on every compute() call — consider module-level constant
**Severity:** info | **Category:** maintainability | **File:** `imagelab-backend/app/operators/geometric/resize_image.py:31`

The `interpolation_map` dictionary is reconstructed on every invocation of `compute()`. While the performance impact is negligible for typical use, the constant mapping of strings to OpenCV flags does not change between calls and would be clearer and marginally more efficient as a module-level or class-level constant.

**Suggestion:**
```
```python
_INTERPOLATION_MAP: dict[str, int] = {
    "LINEAR":  cv2.INTER_LINEAR,
    "AREA":    cv2.INTER_AREA,
    "CUBIC":   cv2.INTER_CUBIC,
    "NEAREST": cv2.INTER_NEAREST,
    "LANCZOS4": cv2.INTER_LANCZOS4,
}

class ResizeImage(BaseOperator):
        ...
        interpolation_flag = _INTERPOLATION_MAP.get(interpolation_method_str, cv2.INTER_LINEAR)
```
```

---

PR #85, opened 2026-02-27
Feature: contour detection

chetanbarakki:
## Description
This PR improves the Contour Detection operator by making it adaptive to different image types (dark background, bright background)

Fixes #50 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Manually tested

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots (if applicable)
White background
image

Nature background
image

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [x] Tests pass locally

PR #87, opened 2026-02-27
feat(filtering): add Contour Detection operator #50

### LINKED ISSUE #50: Add Contour Detection Operator
**Difficulty:** Hard
**Area:** backend, frontend

## Description

Contour detection is a fundamental technique in computer vision used for shape analysis, object detection, and image segmentation. Contours are curves that join all continuous points along a boundary that share the same color or intensity. OpenCV provides robust contour detection through `cv2.findContours()`, which extracts contour points, and `cv2.drawContours()`, which renders them onto an image.

This operator involves a multi-step pipeline within a single compute call: converting to grayscale, applying a binary threshold to create a mask, finding contours on the mask, and then drawing the detected contours onto the original color image. Multiple parameters control the behavior: the retrieval mode determines the contour hierarchy (external only vs. full tree), the approximation method controls point density, and color/thickness control the visual rendering.

## Acceptance Criteria

- [ ] A contour detection operator is implemented in the backend following the existing operator pattern
- [ ] The operator detects contours and draws them on the original image
- [ ] Retrieval mode parameter supports at least EXTERNAL and TREE options
- [ ] Approximation method parameter supports SIMPLE and NONE
- [ ] Color and thickness parameters control the appearance of drawn contours
- [ ] The operator is registered and available in the processing pipeline
- [ ] A Blockly block exists with dropdowns for mode and method, a color picker, and a thickness field, appearing in the filtering category
- [ ] End-to-end: the operator visibly draws contour outlines on an image with distinct shapes

## Resources

- [OpenCV Contours tutorial](https://docs.opencv.org/4.x/d4/d73/tutorial_py_contours_begin.html)
- [cv2.findContours](https://docs.opencv.org/4.x/d3/dc0/group__imgproc__shape.html#gadf1ad6a0b82947fa1fe3c3d497f260e0)
- [cv2.drawContours](https://docs.opencv.org/4.x/d6/d6e/group__imgproc__draw.html#ga746c0625f1781f1ffc9056259103edbc)
- Existing color utility: `imagelab-backend/app/utils/color.py`

Jayant-kernel:
## Description
Implements Issue #50 natively mapping Contour Extractor tools. Added `filtering_contourdetection` backend operators utilizing robust `cv2.findContours` binary thresholds supporting `EXTERNAL`/`TREE` hierarchy modes. Extended frontend Blockly logic enabling parameters directly for custom strokes.

Fixes #50

## Type of Change
- [ ] New feature

## Checklist
- [x] My code follows the project's style guidelines (Ruff Formatted)
- [x] I have generated unit tests mapping logical structural matrices
- [x] Tests pass locally (Pytest & npm build)

##What:
- Adds ContourDetection operator (mode: EXTERNAL/TREE, method: SIMPLE/NONE, color, thickness)
- Registers operator in backend registry
- Adds frontend Blockly block + category entry
- Adds pytest regression tests (no-contours + grayscale + BGRA)

## How tested
Backend:
- `cd imagelab-backend && uv run ruff format . && uv run ruff check .`
- `cd imagelab-backend && uv run pytest -v`

Frontend:
- `cd imagelab-frontend && npm run lint`
- `cd imagelab-frontend && npm run build`

## Notes
- Operator is **un-opinionated**: does not force thresholding/blur; expects a mask or single-channel image for `findContours`.
- Handles grayscale and BGRA inputs without crashing.

## Checklist
- [x] Focused diff (only contour-related files)
- [x] ruff format + ruff check pass
- [x] pytest pass
- [x] frontend lint/build pass

### DIFF
--- a/imagelab-backend/app/operators/filtering/contour_detection.py
+++ b/imagelab-backend/app/operators/filtering/contour_detection.py
@@ -4,63 +4,43 @@ import numpy as np
 from app.operators.base import BaseOperator
 from app.utils.color import hex_to_bgr
 
-MODE_MAP = {"EXTERNAL": cv2.RETR_EXTERNAL, "TREE": cv2.RETR_TREE}
-METHOD_MAP = {"SIMPLE": cv2.CHAIN_APPROX_SIMPLE, "NONE": cv2.CHAIN_APPROX_NONE}
+_MODE_MAP = {"EXTERNAL": cv2.RETR_EXTERNAL, "TREE": cv2.RETR_TREE}
+_METHOD_MAP = {"SIMPLE": cv2.CHAIN_APPROX_SIMPLE, "NONE": cv2.CHAIN_APPROX_NONE}
 
 
 class ContourDetection(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        """
-        Detect contours in a single-channel (binary/grayscale) image and draw
-        them onto a copy of the input.
-
-        Params:
-            mode (str): "EXTERNAL" (outermost contours only) or "TREE" (full hierarchy).
-            method (str): "SIMPLE" (compressed) or "NONE" (all contour points).
-            rgbcolors_input (str): Hex color string for the drawn contours (e.g. "#00ff00").
-            thickness (int): Stroke width in pixels (>= 1, <= 50).
-
-        Note:
-            Float images are scaled to [0, 255] before processing.
-            Grayscale input is promoted to BGR in the returned image.
-            Does not apply thresholding or blurring; the caller is responsible
-            for ensuring the input is a suitable binary or edge mask.
-        """
         mode_str = str(self.params.get("mode", "EXTERNAL")).upper()
         method_str = str(self.params.get("method", "SIMPLE")).upper()
         thickness = int(self.params.get("thickness", 2))
-        hex_color = self.params.get("rgbcolors_input", "#00ff00")
-        bgr_color = hex_to_bgr(hex_color)
+        bgr_color = hex_to_bgr(self.params.get("rgbcolors_input", "#00ff00"))
 
-        if mode_str not in MODE_MAP:
-            raise ValueError(f"Invalid contour mode '{mode_str}'. Must be one of {list(MODE_MAP)}.")
-        if method_str not in METHOD_MAP:
-            raise ValueError(f"Invalid contour method '{method_str}'. Must be one of {list(METHOD_MAP)}.")
-        mode = MODE_MAP[mode_str]
-        method = METHOD_MAP[method_str]
+        if mode_str not in _MODE_MAP:
+            raise ValueError(f"Invalid contour mode '{mode_str}'. Must be one of {list(_MODE_MAP)}.")
+        if method_str not in _METHOD_MAP:
+            raise ValueError(f"Invalid contour method '{method_str}'. Must be one of {list(_METHOD_MAP)}.")
+        if thickness < 1 or thickness > 50:
+            raise ValueError(f"thickness must be between 1 and 50, got {thickness}")
 
-        if thickness < 1:
-            raise ValueError(f"thickness must be >= 1, got {thickness}")
-        if thickness > 50:
-            raise ValueError(f"thickness must be <= 50, got {thickness}")
+        mode = _MODE_MAP[mode_str]
+        method = _METHOD_MAP[method_str]
 
-        if image.ndim == 3 and image.shape[2] == 3:
+        # Convert to single-channel for findContours
+        if len(image.shape) == 3 and image.shape[2] == 3:
             gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
-        elif image.ndim == 3 and image.shape[2] == 4:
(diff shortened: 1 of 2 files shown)


Jayant-kernel:
Hi @ivantha! Just pushed an update to my PR. I noticed the discussion about adding automatic blurring/inversion/otsu inside the contour operator.

My PR deliberately avoids doing that to preserve the Single Responsibility Principle. In a pipeline editor like ImageLab, I believe the Contour operator should strictly just find and draw contours. If users want to reduce noise or apply Otsu thresholding first, they can simply snap the existing Blur and Threshold blocks before the Contour block in their workspace! This keeps the operator modular and predictable.

I did, however, implement the BGRA/transparency safety check so 4-channel images don't crash the pipeline. Let me know which architectural direction you prefer!

Jayant-kernel:
@ivantha 
I noticed #85 is also addressing #50. In #87 I kept the diff very small (5 files, 1 commit) and added pytest coverage (no-contours + grayscale + BGRA) along with the frontend block wiring. 

If you’d rather proceed with #85, I’m totally happy to close #87 and help there — just let me know your preference.

imagelab-backend/app/operators/filtering/contour_detection.py
```diff
@@ -0,0 +1,53 @@
+import cv2
+import numpy as np
+
+from app.operators.base import BaseOperator
+from app.utils.color import hex_to_bgr
+
+
+class ContourDetection(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
```
ivantha: **[LOW]** Missing docstring on `compute()` method

The `compute()` method has no docstring, making it harder for users of the operator to understand:
- What input format is expected (`uint8` only? Any dtype? Any channel count?).
- What parameters are supported and their allowed values.
- What the output format will be (notably, that grayscale input may be converted to BGR).
- Any known limitations or caveats.

Other operators in the codebase may follow a similar pattern, but given this operator's non-trivial output format behavior (described above), documentation is especially important here.

**Suggestion:**
```
Add a docstring:

```python
    """
    Detect contours in a single-channel (binary/grayscale) image and draw
    them onto a copy of the input.

    Params:
        mode (str): "EXTERNAL" (outermost contours only) or "TREE" (full hierarchy).
        method (str): "SIMPLE" (compressed) or "NONE" (all contour points).
        rgbcolors_input (str): Hex color string for the drawn contours (e.g. "#00ff00").
        thickness (int): Stroke width in pixels (>= 1).

    Note:
        Expects a uint8 image. Float images are scaled to [0,255].
        Grayscale input is promoted to BGR in the returned image.
        Does not apply thresholding or blurring; the caller is responsible
        for ensuring the input is a suitable binary or edge mask.
    """
```
```

File: imagelab-frontend/src/blocks/definitions/filtering.blocks.ts [hunk 2]
```diff
@@ -112,4 +112,35 @@ export const filteringBlocks = [
...
+      { type: "input_dummy" },
+      { type: "field_colour", name: "rgbcolors_input", colour: "#00ff00" },
+      { type: "input_dummy" },
+      { type: "field_number", name: "thickness", value: 2, min: 1 },
+    ],
+    previousStatement: null,
+    nextStatement: null,
+    style: "filtering_style",
+    tooltip: "Detects contours on an image and renders them over the original graphic.",
+  },
```
ivantha: **[LOW]** No maximum thickness constraint in frontend block definition

The `field_number` for `thickness` sets `min: 1` but has no `max`. A user can enter arbitrarily large values (e.g., `9999`) in the Blockly UI. While the backend can enforce limits, a maximum on the frontend improves UX by preventing obviously invalid inputs at the point of entry and gives users a clear indication of the valid range.

**Suggestion:**
```
Add a `max` constraint consistent with what the backend will enforce:

```typescript
{ type: "field_number", name: "thickness", value: 2, min: 1, max: 50 },
```

Choose a maximum that makes semantic sense for the use case (e.g., 50 pixels is a very thick contour stroke for most image sizes).
```

File: imagelab-backend/app/operators/registry.py [hunk 3]
```diff
@@ -85,6 +86,7 @@
     "filtering_erosion": Erosion,
     "filtering_dilation": Dilation,
     "filtering_morphological": Morphological,
+    "filtering_contourdetection": ContourDetection,
```
ivantha: **[LOW]** Contour detection may be miscategorized as a "filtering" operation

Contour detection is fundamentally an **analysis/drawing** operation—it detects structural features and renders geometry onto an image. The existing `filtering` category groups operations like `BilateralFilter`, `Erosion`, and `BoxFilter`, which transform pixel values. Placing contour detection here may confuse users who expect filtering to modify pixel intensities rather than draw geometric shapes.

This is a design discussion point rather than a defect, but worth raising before the category structure grows harder to change.

**Suggestion:**
```
Consider placing `ContourDetection` under an existing `drawing` or `analysis` category if one exists, or creating a new `shape_analysis` category. At minimum, open a follow-up issue to discuss the operator taxonomy. The frontend label `"Contour Detection"` is accurate, but the category label `"Filtering"` shown to the user may be misleading.
```

Jayant-kernel:
@ivantha 
please rereview it

Issue #59, opened 2026-02-25
Backend API Integration Tests


ivantha:
**Difficulty:** Hard
**Area:** backend, testing

## Description

While unit tests for individual operators verify that each computation is correct, integration tests for the `/api/pipeline/execute` endpoint verify that the entire request-response cycle works correctly: JSON parsing, base64 image decoding, pipeline step execution, result encoding, and error response formatting. These tests catch issues at the API boundary that unit tests miss, such as incorrect request/response schemas, HTTP status codes, and serialization problems.

Integration tests should use FastAPI's `TestClient` to make real HTTP requests to the application without needing a running server. This approach tests the full stack from HTTP layer through to the operators and back.

## Acceptance Criteria

- [ ] Integration tests exist for the pipeline execution API endpoint
- [ ] Tests make real HTTP requests against the application (using a test client)
- [ ] Tests cover: successful single-step, multi-step, empty pipeline, unknown operator, invalid base64, large pipeline, and missing fields
- [ ] A reusable base64 image fixture is provided
- [ ] Tests verify correct HTTP status codes and response structure
- [ ] Tests verify that error responses contain meaningful error messages
- [ ] All tests pass with `uv run pytest -v`
- [ ] Tests are independent and do not require external services (database, file system)

## Resources

- [FastAPI TestClient documentation](https://fastapi.tiangolo.com/tutorial/testing/)
- [Starlette TestClient](https://www.starlette.io/testclient/)
- API endpoint: `imagelab-backend/app/main.py` or route file
- Pipeline models: `imagelab-backend/app/models/pipeline.py`

Issue #48, opened 2026-02-25
Backend Unit Tests for Pipeline Executor


ivantha:
**Difficulty:** Hard
**Area:** backend, testing

## Description

The pipeline executor (`imagelab-backend/app/services/pipeline_executor.py`) is the central orchestration component that takes a list of pipeline steps, looks up each operator in the registry, and sequentially applies them to an image. Despite being the most critical backend service, it currently lacks unit test coverage.

Testing the executor is more complex than testing individual operators because it involves the registry lookup mechanism, sequential step execution, error propagation with step indices, and handling of edge cases like empty pipelines and unknown operator types. These tests are essential for ensuring pipeline reliability and providing clear error messages when something goes wrong.

## Acceptance Criteria

- [ ] Tests exist for the pipeline executor service
- [ ] Tests cover: empty pipeline, single-step pipeline, multi-step pipeline, unknown operator type, and error propagation with step index
- [ ] A reusable base64 image fixture is provided
- [ ] Tests use real operators (not mocks) for positive test cases
- [ ] Error test cases verify that the response contains meaningful error information
- [ ] All tests pass with `uv run pytest -v`
- [ ] Tests are independent and can run in any order

## Resources

- [pytest documentation](https://docs.pytest.org/en/stable/)
- Pipeline executor: `imagelab-backend/app/services/pipeline_executor.py`
- Pipeline models: `imagelab-backend/app/models/pipeline.py`
- Image utilities: `imagelab-backend/app/utils/image.py`
- Operator registry: `imagelab-backend/app/operators/registry.py`

areebahmeddd:
Hey, i would like to pick this up? 🙂

My approach is: set up a shared base64 image fixture in conftest, then write tests covering empty pipeline, NOOP skipping, single/multi-step with real operators, unknown operator type and error propagation

Jayant-kernel:
@ivantha 
I am working on this issue and will submit a PR

areebahmeddd:
@Jayant-kernel hey, i'm in the middle of implementing this too. would you mind taking a look at other issues? there's no comment or pr too! 🙂

Issue #47, opened 2026-02-25
[Chore] Add unit tests for the ReflectImage geometric operator


ivantha:
**Difficulty:** Easy
**Area:** backend

## Description
Already landed: three of the four geometric operators now have coverage. Affine is covered by `imagelab-backend/tests/operators/geometric/test_affine_image.py`, scale by `imagelab-backend/tests/operators/geometric/test_scale_image.py` plus `imagelab-backend/tests/test_scale_image.py`, and rotate by `imagelab-backend/tests/test_rotate_image.py`. What survives is `ReflectImage` (`imagelab-backend/app/operators/geometric/reflect_image.py:13`), which no test in the repository references.

`ReflectImage` maps a `type` param of "X", "Y" or "Both" onto `cv2.flip` codes 0, 1 and -1. Axis naming is exactly the kind of convention that inverts silently under refactoring, and a shape-only assertion would not catch it, so the tests need pixel-level assertions on where a known bright pixel lands.

## Acceptance Criteria
- [ ] `imagelab-backend/tests/operators/geometric/test_reflect_image.py` exists and follows the single-bright-pixel style of the neighbouring `test_affine_image.py`
- [ ] A test pins the default (no `type` param) to "X", flipping around the x-axis, matching `FLIP_CODES` in `reflect_image.py:6-10`
- [ ] Separate tests pin "X", "Y" and "Both" by asserting the destination coordinates of a known bright pixel, not only that the shape is unchanged
- [ ] A test asserts an unrecognised `type` value falls back to flip code 0 rather than raising
- [ ] A test uses a non-square image so a transposed axis cannot pass silently
- [ ] `uv run pytest -v` passes from `imagelab-backend/`

## Resources
- `imagelab-backend/app/operators/geometric/reflect_image.py:6-17` - `FLIP_CODES` and the `compute` under test
- `imagelab-backend/tests/operators/geometric/test_affine_image.py:6-9` - the `_single_bright_pixel` helper worth mirroring
- `imagelab-backend/tests/conftest.py` - shared fixtures
- [cv2.flip reference](https://docs.opencv.org/4.x/d2/de8/group__core__array.html#gaca7be533e3dac7feb70fc60635adf441)
- [pytest documentation](https://docs.pytest.org/en/stable/)

PR #96, opened 2026-02-28
add unit + integration tests for imagelab-be

### LINKED ISSUE #46: Backend Unit Tests for Blurring Operators
**Difficulty:** Medium
**Area:** backend, testing

## Description

The three blurring operators (`blur.py`, `gaussian_blur.py`, `median_blur.py`) currently have no unit test coverage. These operators are fundamental image processing functions and should be thoroughly tested to ensure they produce correct results, handle edge cases gracefully, and maintain expected behavior as the codebase evolves.

Tests should cover correct output shape preservation, parameter handling (defaults and explicit values), the odd-kernel-size enforcement for Gaussian and median blur, and behavior with both grayscale and color input images. Using numpy arrays as test fixtures avoids the need for actual image files and keeps tests fast and self-contained.

## Acceptance Criteria

- [ ] Tests exist for all three blurring operators (blur, Gaussian blur, median blur)
- [ ] Test fixtures provide reusable sample images (color and grayscale)
- [ ] Each operator has at least 5 test cases covering: default parameters, custom parameters, output shape, even-to-odd kernel correction, and large kernels
- [ ] All tests pass with `uv run pytest -v`
- [ ] Tests run quickly without file I/O (using in-memory arrays)
- [ ] Tests follow pytest conventions and use fixtures for shared setup

## Resources

- [pytest documentation](https://docs.pytest.org/en/stable/)
- [pytest fixtures](https://docs.pytest.org/en/stable/how-to/fixtures.html)
- Operator implementations: `imagelab-backend/app/operators/blurring/`

areebahmeddd:
## Description

- Restructures the entire `imagelab-backend/tests/` directory following pytest best practices
- Adds end to end unit and integration test coverage for all 44 operators, the pipeline executor, and the REST API
- Configures `pytest-cov` with `--cov=app --cov-report=term-missing` and HTML output via `pyproject.toml`

Fixes #46, #47, #48, #59

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] New tests added
- [x] Existing tests pass
- [x] Manual testing

**Results:** 311 tests, 96% coverage, ruff clean.

image

## Checklist

- [x] Tests pass locally


ivantha (commented):
## ◎ Argus Review

This PR adds a comprehensive, well-structured test suite (311 tests, 96% coverage) for imagelab-backend. The test directory is cleanly organized into unit and integration layers with sensible shared fixtures. There are two notable issues that warrant attention: a duplicate `pytest-cov` dependency declaration with conflicting version constraints in `pyproject.toml`, and a test logic bug where `assert resp.error != ""` silently passes even when `error` is `None`. Beyond these, several tests have correctness or completeness gaps worth addressing before merge.

imagelab-backend/tests/unit/operators/test_blurring.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_uniform_image_unchanged(self):
+        uniform = np.full((50, 50, 3), 128, dtype=np.uint8)
+        np.testing.assert_array_equal(Blur({"widthSize": "5", "heightSize": "5"}).compute(uniform), uniform)
+
+
+class TestGaussianBlur:
+    def test_color_shape_preserved(self, color_image):
+        assert GaussianBlur({}).compute(color_image).shape == color_image.shape
+
+    def test_gray_shape_preserved(self, gray_image):
```
ivantha: **[INFO]** Positive: Statistical and mathematical property tests are well-designed

Tests like `test_large_kernel_reduces_variance` and `test_uniform_image_unchanged` go beyond shape/dtype checks to verify *behavioral* correctness (blurring reduces variance; uniform images are unaffected). Similarly, `test_uniform_image_produces_zero` in `TestLaplacian` correctly validates a mathematical property of the Laplacian operator. These property-based tests provide much stronger guarantees than pure shape checks and are a highlight of this PR.

**Suggestion:**
```
This pattern should be extended to more operators — for example, verifying that `Erosion` + `Dilation` with the same kernel produces a result within a bounded distance of the original (opening/closing idempotence), or that `ScaleImage(fx=2)` followed by `ScaleImage(fx=0.5)` approximately recovers the original.
```

imagelab-backend/pyproject.toml
```diff
@@ -33,10 +34,24 @@ quote-style = "double"
...
 [tool.hatch.build.targets.wheel]
 packages = ["app"]
 
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
+
+[dependency-groups]
+dev = [
+    "pytest-cov>=7.0.0",
```
ivantha: **[HIGH]** Duplicate pytest-cov dependency with conflicting version constraints

`pytest-cov` is declared in two separate sections with different minimum versions:

- `[project.optional-dependencies]` dev: `"pytest-cov>=5.0"` (line 19)
- `[dependency-groups]` dev: `"pytest-cov>=7.0.0"` (lines 53–56)

`[dependency-groups]` is a PEP 735 / uv-specific concept, while `[project.optional-dependencies]` is standard PEP 508. Declaring the same package in both with different floors is confusing and tool-dependent:
- `pip install .[dev]` will honour only `>=5.0`.
- `uv sync --group dev` will intersect both and resolve `>=7.0.0`.

The inconsistency can also cause subtle CI failures if different tools are used in different environments.

**Suggestion:**
```
Pick one mechanism. Since the project is already using uv, consolidate under `[dependency-groups]` with the stricter bound and remove the entry from `[project.optional-dependencies]`:
```toml
[project.optional-dependencies]
dev = [
    "pytest>=8.0",
    # pytest-cov moved to [dependency-groups]
    "ruff>=0.8",
    ...
]

[dependency-groups]
dev = [
    "pytest-cov>=7.0.0",
]
```
Alternatively, keep only `[project.optional-dependencies]` and bump its floor to `>=7.0.0`.
```

File: imagelab-backend/tests/integration/test_pipeline_executor.py [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+        assert resp.step == 0
+
+    def test_failure_has_nonempty_error_message(self, base64_png_image):
+        req = PipelineRequest(
+            image=base64_png_image,
+            pipeline=[PipelineStep(type="does_not_exist")],
+        )
+        resp = execute_pipeline(req)
+        assert resp.error != ""
+        assert resp.image is None
```
ivantha: **[HIGH]** Test assertion passes silently when error is None (logic bug)

In `test_failure_has_nonempty_error_message`:
```python
assert resp.error != ""
assert resp.image is None
```
In Python, `None != ""` evaluates to `True`, so the first assertion passes even when `resp.error` is `None`. This means the test would not catch a regression where the executor returns a failure response without populating the error message at all. The intent is clearly to assert a non-empty string is present, but the assertion is broken.

**Suggestion:**
```
Use an explicit non-None and non-empty check:
```python
assert resp.error is not None, "error field should not be None on failure"
assert resp.error != "", "error message should not be empty string"
assert resp.image is None
```
Or more concisely:
```python
assert resp.error, "expected a non-empty error message"
assert resp.image is None
```
```

File: imagelab-backend/tests/integration/test_api.py [hunk 4]
```diff
@@ -0,0 +1,172 @@
+import pytest
+from fastapi.testclient import TestClient
+
+from app.main import app
+
+
+@pytest.fixture(scope="module")
+def client():
+    with TestClient(app, raise_server_exceptions=False) as c:
+        yield c
```
ivantha: **[MEDIUM]** raise_server_exceptions=False silently swallows 500 errors

The module-level `client` fixture suppresses server exceptions:
```python
with TestClient(app, raise_server_exceptions=False) as c:
```
This means if the application raises an unhandled exception (e.g., a bug in the executor or a broken import), FastAPI will emit a 500 response and the test will attempt to parse the error JSON. Many assertions like `resp.status_code == 200` or `resp.json()["success"] is True` will then produce confusing failures (e.g., JSONDecodeError or KeyError) rather than the actual traceback, making regressions extremely hard to diagnose.

**Suggestion:**
```
Set `raise_server_exceptions=True` (the default). If specific tests intentionally exercise 500 paths, create a separate fixture with exception suppression scoped to only those tests:
```python
@pytest.fixture(scope="module")
def client():
    with TestClient(app) as c:  # raise_server_exceptions=True by default
        yield c
```
```

File: imagelab-backend/tests/integration/test_api.py [hunk 5]
```diff
@@ -0,0 +1,172 @@
...
+    def test_response_has_required_fields(self, client, base64_png_image):
+        resp = client.post(
+            "/api/pipeline/execute",
+            json={"image": base64_png_image, "pipeline": []},
+        )
+        data = resp.json()
+        assert "success" in data
+
+    def test_success_response_has_image(self, client, base64_png_image):
+        resp = client.post(
```
ivantha: **[MEDIUM]** test_response_has_required_fields only checks for key presence, not value

`test_response_has_required_fields` only asserts `"success" in data`, which passes even if `success` is `None`, `0`, or any falsy value. The test name implies more thorough validation than what is actually performed. It also only checks one key (`success`) while the PR description references a richer schema with `image`, `image_format`, and `error` fields.

**Suggestion:**
```
Expand the assertion to validate the expected response schema more rigorously:
```python
def test_response_has_required_fields(self, client, base64_png_image):
    resp = client.post(
        "/api/pipeline/execute",
        json={"image": base64_png_image, "pipeline": []},
    )
    data = resp.json()
    assert "success" in data
    assert isinstance(data["success"], bool)
    assert "image" in data
    assert "image_format" in data
```
```

File: imagelab-backend/pyproject.toml [hunk 6]
```diff
@@ -33,10 +34,24 @@ quote-style = "double"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
```
ivantha: **[MEDIUM]** HTML coverage report generated on every test run via addopts

The `addopts` setting unconditionally runs `--cov-report=html:htmlcov` on every test invocation:
```toml
addopts = "--cov=app --cov-report=term-missing --cov-report=html:htmlcov -q"
```
This means every developer running tests locally will write an `htmlcov/` directory, which:
1. Slows down test runs.
2. Can create unexpected dirty working-tree state.
3. May be committed accidentally if `htmlcov/` is not in `.gitignore`.

Additionally, the `-q` (quiet) flag suppresses test names in output, which makes debugging individual failures harder.

**Suggestion:**
```
Keep `--cov` and `--cov-report=term-missing` in `addopts` for fast feedback; move HTML generation to a CI-only step:
```toml
# pyproject.toml
[tool.pytest.ini_options]
testpaths = ["tests"]
addopts = "--cov=app --cov-report=term-missing"
```
```yaml
# CI step
- run: pytest --cov-report=html:htmlcov
```
```

File: imagelab-backend/pyproject.toml [hunk 7]
```diff
@@ -33,10 +34,24 @@ quote-style = "double"
...
+addopts = "--cov=app --cov-report=term-missing --cov-report=html:htmlcov -q"
+
+[tool.coverage.run]
+source = ["app"]
```
ivantha: **[MEDIUM]** Coverage silently omits app/operators/basic/* without explanation

The coverage omit list includes `app/operators/basic/*`:
```toml
omit = ["app/database.py", "app/__init__.py", "app/operators/__init__.py", "app/operators/basic/*"]
```
While `database.py` and `__init__.py` are reasonable omissions, silently excluding an entire operator category from coverage reporting is concerning. The `basic_readimage` and `basic_writeimage` operators are exercised indirectly in pipeline tests as "noop" blocks, and their exclusion means regressions in those operators would not be flagged by the coverage gate.

**Suggestion:**
```
Add an inline comment explaining why basic operators are excluded, and verify this is intentional:
```toml
omit = [
    "app/database.py",        # DB migration scaffolding, not business logic
    "app/__init__.py",
    "app/operators/__init__.py",
    "app/operators/basic/*",  # Read/write stubs, intentionally excluded (noop at runtime)
]
```
Alternatively, if these are genuinely testable, add unit tests and remove the omission.
```

File: imagelab-backend/uv.lock [hunk 8]
```diff
@@ -156,7 +237,6 @@ wheels = [
...
     { url = "https://files.pythonhosted.org/packages/f5/2c/d2d506ebd8abcb57386ec4f7ba20f4030cbe56eae541bc6fd6ef399c0b41/greenlet-3.3.2-cp314-cp314t-manylinux_2_24_ppc64le.manylinux_2_28_ppc64le.whl", hash = "sha256:b8bddc5b73c9720bea487b3bffdb1840fe4e3656fba3bd40aa1489e9f37877ff", size = 658225, upload-time = "2026-02-20T20:56:02.527Z" },
-    { url = "https://files.pythonhosted.org/packages/d1/67/8197b7e7e602150938049d8e7f30de1660cfb87e4c8ee349b42b67bdb2e1/greenlet-3.3.2-cp314-cp314t-manylinux_2_24_s390x.manylinux_2_28_s390x.whl", hash = "sha256:59b3e2c40f6706b05a9cd299c836c6aa2378cabe25d021acd80f13abf81181cf", size = 666581, upload-time = "2026-02-20T21:02:51.526Z" },
     { url = "https://files.pythonhosted.org/packages/8e/30/3a09155fbf728673a1dea713572d2d31159f824a37c22da82127056c44e4/greenlet-3.3.2-cp314-cp314t-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl", hash = "sha256:b26b0f4428b871a751968285a1ac9648944cea09807177ac639b030bddebcea4", size = 657907, upload-time = "2026-02-20T20:21:05.259Z" },
```
ivantha: **[LOW]** s390x greenlet wheel entries manually removed from uv.lock

Three entries for the IBM Z (s390x) architecture have been removed from `uv.lock`:
- `greenlet-3.3.2-cp312-cp312-manylinux_2_24_s390x...`
- `greenlet-3.3.2-cp313-cp313-manylinux_2_24_s390x...`
- `greenlet-3.3.2-cp314-cp314-manylinux_2_24_s390x...`

Lock files should generally be auto-generated by running `uv lock`, not manually edited. Selective removal of architecture-specific wheel entries is unusual and suggests either a manual edit (fragile) or a regeneration on a system that excluded s390x support. If the CI ever runs on s390x, installation would fail.

**Suggestion:**
```
Regenerate the lock file cleanly with `uv lock` on a representative developer machine rather than manually editing wheel entries. If s390x is intentionally unsupported, document this in a `[tool.uv]` constraint or in the README rather than by lock file surgery.
```

imagelab-backend/tests/integration/test_api.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+        assert resp.json()["image_format"] == "png"
+
+    def test_noop_blocks_skipped(self, client, base64_png_image):
+        resp = client.post(
+            "/api/pipeline/execute",
+            json={
+                "image": base64_png_image,
+                "pipeline": [
+                    {"type": "basic_readimage"},
```
ivantha: **[INFO]** Noop operator tests do not verify output image is unchanged

`test_noop_blocks_skipped` verifies that passing `basic_readimage` and `basic_writeimage` into the pipeline returns `success: True`, but does not check that the output image is equivalent to the input. A regression where these operators corrupt the image would not be caught.

**Suggestion:**
```
Decode both the input and output images and compare them:
```python
def test_noop_blocks_skipped(self, client, base64_png_image):
    resp = client.post(
        "/api/pipeline/execute",
        json={
            "image": base64_png_image,
            "pipeline": [
                {"type": "basic_readimage"},
                {"type": "basic_writeimage"},
            ],
        },
    )
    data = resp.json()
    assert data["success"] is True
    # Noop operators should not modify the image
    assert data["image"] == base64_png_image
```
```

File: imagelab-backend/tests/conftest.py [hunk 10]
```diff
@@ -0,0 +1,41 @@
...
+    img[50:150, 50:150] = [80, 120, 200]
+    return img
+
+
+@pytest.fixture
+def base64_png_image() -> str:
+    img = np.zeros((100, 100, 3), dtype=np.uint8)
+    img[25:75, 25:75] = [0, 200, 100]
+    _, buf = cv2.imencode(".png", img)
+    return base64.b64encode(buf).decode("utf-8")
```
ivantha: **[INFO]** Positive: Well-designed shared fixtures covering key image types

The `conftest.py` provides a clean set of reusable fixtures that cover the essential image variants (color, gray, binary, large, base64-encoded). Using `np.random.default_rng` with fixed seeds in individual tests rather than fixtures keeps tests deterministic without coupling fixtures to randomness. The `base64_png_image` fixture correctly uses `cv2.imencode` to produce a realistic round-trippable payload rather than a synthetic string.

**Suggestion:**
```
No change needed. Consider adding an RGBA fixture to enable broader testing of RGBA-to-BGR conversion paths (currently tested inline in `TestBilateralFilter`).
```

PR #97, opened 2026-02-28
feat: add keyboard shortcut hint to Run Pipeline toolbar tooltip (#58)

### LINKED ISSUE #58: Add Keyboard Shortcuts for Common Actions
**Difficulty:** Medium
**Area:** frontend

## Description

ImageLab currently requires mouse interaction for all toolbar actions (run pipeline, download, undo, redo). Adding keyboard shortcuts for these frequently used actions will significantly improve the user experience, especially for power users who are building and testing pipelines iteratively. Keyboard shortcuts are a standard feature in any editor-style application.

While Blockly has its own built-in undo/redo shortcuts within the workspace, these do not work when focus is outside the Blockly canvas. Global keyboard shortcuts should work regardless of which component has focus, and shortcut hints should be displayed in toolbar button tooltips so users can discover them.

## Acceptance Criteria

- [ ] A keyboard shortcut handler is implemented for all planned shortcuts
- [ ] `Ctrl+Enter` / `Cmd+Enter` triggers pipeline execution
- [ ] `Ctrl+S` / `Cmd+S` downloads the processed image and prevents browser default save
- [ ] `Ctrl+Z` / `Cmd+Z` performs undo
- [ ] `Ctrl+Y` / `Cmd+Y` or `Ctrl+Shift+Z` performs redo
- [ ] `Delete` / `Backspace` removes the selected Blockly block (when not in an input field)
- [ ] Shortcuts work regardless of which component has focus
- [ ] Shortcuts do not interfere with typing in Blockly field editors (number inputs, text fields)
- [ ] Toolbar button tooltips display the corresponding keyboard shortcuts
- [ ] Both `Ctrl` (Windows/Linux) and `Cmd` (macOS) modifier keys are supported

## Resources

- [MDN KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)
- [Blockly selected block API](https://developers.google.com/blockly/reference/js/blockly.getselected)
- Current Toolbar: `imagelab-frontend/src/components/Toolbar.tsx`
- Current Layout: `imagelab-frontend/src/components/Layout.tsx`

Anirudh-x:
## Description

Added the missing keyboard shortcut hint to the **Run Pipeline** toolbar button tooltip. All other toolbar buttons (Download, Undo, Redo) already displayed their shortcuts in tooltips, but the Run button only showed `"Run Pipeline"` with no shortcut hint. This change makes the tooltip consistent with the rest of the toolbar by showing `Run Pipeline (Ctrl+Enter)` on Windows/Linux and `Run Pipeline (⌘Enter)` on macOS.

Fixes #58

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Hovered over the Run Pipeline button on Windows — tooltip shows `Run Pipeline (Ctrl+Enter)`
- Verified the `isMac` / `mod` variable correctly switches to `⌘Enter` on macOS
- Confirmed `Ctrl+Enter` still triggers pipeline execution as expected
- Verified no regression in other toolbar button tooltips

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-frontend/src/components/Toolbar.tsx
+++ b/imagelab-frontend/src/components/Toolbar.tsx
@@ -4,17 +4,14 @@ import { usePipelineStore } from "../store/pipelineStore";
 import { executePipeline } from "../api/pipeline";
 import { extractPipeline } from "../hooks/usePipeline";
 import { useKeyboardShortcuts } from "../hooks/useKeyboardShortcuts";
+import { useModKey } from "../hooks/useModKey";
 
 interface ToolbarProps {
   workspace: Blockly.WorkspaceSvg | null;
 }
 
-// Detect macOS to show Cmd vs Ctrl in tooltips
-const isMac =
-  typeof navigator !== "undefined" && /mac/i.test(navigator.platform || navigator.userAgent);
-const mod = isMac ? "⌘" : "Ctrl+";
-
 export default function Toolbar({ workspace }: ToolbarProps) {
+  const mod = useModKey();
   const {
     originalImage,
     imageFormat,
@@ -100,7 +97,7 @@ export default function Toolbar({ workspace }: ToolbarProps) {
 
   return (
     <div className="h-10 flex items-center gap-1 px-3 bg-white border-b border-gray-200 flex-shrink-0">
-      <button onClick={handleNew} className={iconBtn} title="New">
+      <button onClick={handleNew} className={iconBtn} title="New" aria-label="New">
         <FilePlus size={18} />
       </button>
       <button
@@ -108,16 +105,17 @@ export default function Toolbar({ workspace }: ToolbarProps) {
         disabled={!processedImage}
         className={iconBtn}
         title={`Download (${mod}S)`}
+        aria-label={`Download (${mod}S)`}
       >
         <Download size={18} />
       </button>
 
       <div className={separator} />
 
-      <button onClick={handleUndo} className={iconBtn} title={`Undo (${mod}Z)`}>
+      <button onClick={handleUndo} className={iconBtn} title={`Undo (${mod}Z)`} aria-label={`Undo (${mod}Z)`}>
         <Undo2 size={18} />
       </button>
-      <button onClick={handleRedo} className={iconBtn} title={`Redo (${mod}Y or ${mod}⇧Z)`}>
+      <button onClick={handleRedo} className={iconBtn} title={`Redo (${mod}Y or ${mod}⇧Z)`} aria-label={`Redo (${mod}Y or ${mod}⇧Z)`}>
         <Redo2 size={18} />
       </button>
 
@@ -128,6 +126,7 @@ export default function Toolbar({ workspace }: ToolbarProps) {
         disabled={isExecuting || !originalImage}
         className="flex items-center gap-1.5 px-3 py-1 rounded-md text-sm font-medium text-white bg-indigo-500 hover:bg-indigo-600 disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
         title={`Run Pipeline (${mod}Enter)`}
+        aria-label={`Run Pipeline (${mod}Enter)`}
       >
         {isExecuting ? <Loader2 size={16} className="animate-spin" /> : <Play size={16} />}
         {isExecuting ? "Running..." : "Run"}
--- a/imagelab-frontend/src/components/Toolbar.tsx
+++ b/imagelab-frontend/src/components/Toolbar.tsx
@@ -144,10 +144,10 @@ export default function Toolbar({ workspace }: ToolbarProps) {
             </span>
             <span
               className={`text-[10px] uppercase font-bold tracking-wide ${complexity === "High"
-                  ? "text-red-500"
-                  : complexity === "Medium"
-                    ? "text-orange-500"
-                    : "text-green-500"
+                ? "text-red-500"
+                : complexity === "Medium"
+                  ? "text-orange-500"
+                  : "text-green-500"
                 }`}
             >
               {complexity} Complexity
--- a/imagelab-frontend/src/components/Toolbar.tsx
+++ b/imagelab-frontend/src/components/Toolbar.tsx
@@ -144,13 +144,12 @@ export default function Toolbar({ workspace }: ToolbarProps) {
               {blockCount} {blockCount === 1 ? "block" : "blocks"}
             </span>
             <span
-              className={`text-[10px] uppercase font-bold tracking-wide ${
-                complexity === "High"
+              className={`text-[10px] uppercase font-bold tracking-wide ${complexity === "High"
                   ? "text-red-500"
                   : complexity === "Medium"
                     ? "text-orange-500"
                     : "text-green-500"
-              }`}
+                }`}
             >
               {complexity} Complexity
             </span>
--- a/imagelab-frontend/src/components/Toolbar.tsx
+++ b/imagelab-frontend/src/components/Toolbar.tsx
@@ -127,7 +127,7 @@ export default function Toolbar({ workspace }: ToolbarProps) {
         onClick={handleRun}
         disabled={isExecuting || !originalImage}
         className="flex items-center gap-1.5 px-3 py-1 rounded-md text-sm font-medium text-white bg-indigo-500 hover:bg-indigo-600 disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
-        title="Run Pipeline"
+        title={`Run Pipeline (${mod}Enter)`}
       >
         {isExecuting ? <Loader2 size={16} className="animate-spin" /> : <Play size={16} />}
         {isExecuting ? "Running..." : "Run"}
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "imagelab",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
--- a/node_modules/.package-lock.json
+++ b/node_modules/.package-lock.json
@@ -1,6 +1,0 @@
-{
-  "name": "imagelab",
-  "lockfileVersion": 3,
-  "requires": true,
-  "packages": {}
-}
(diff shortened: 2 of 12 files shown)


imagelab-frontend/src/components/Toolbar.tsx
```diff
@@ -122,7 +122,7 @@ export default function Toolbar({ workspace }: ToolbarProps) {
         onClick={handleRun}
         disabled={isExecuting || !originalImage}
         className="flex items-center gap-1.5 px-3 py-1 rounded-md text-sm font-medium text-white bg-indigo-500 hover:bg-indigo-600 disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
-        title="Run Pipeline"
+        title={`Run Pipeline (${mod}Enter)`}
       >
         {isExecuting ? <Loader2 size={16} className="animate-spin" /> : <Play size={16} />}
         {isExecuting ? "Running..." : "Run"}
```
ivantha: **[INFO]** Positive: change correctly mirrors the existing tooltip pattern

The modification is minimal, targeted, and follows the same pattern used by the other toolbar buttons (Download, Undo, Redo). Using `title` with a template literal and the shared `mod` variable for platform-conditional text is consistent with the rest of the file. The change is easy to review and carries very low regression risk for the functional pipeline execution path.

**Suggestion:**
```
No action needed. This is a note acknowledging good pattern adherence.
```

imagelab-frontend/src/components/Toolbar.tsx
```diff
@@ -122,7 +122,7 @@ export default function Toolbar({ workspace }: ToolbarProps) {
         onClick={handleRun}
         disabled={isExecuting || !originalImage}
         className="flex items-center gap-1.5 px-3 py-1 rounded-md text-sm font-medium text-white bg-indigo-500 hover:bg-indigo-600 disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
-        title="Run Pipeline"
+        title={`Run Pipeline (${mod}Enter)`}
```
ivantha: **[LOW]** Tooltip strings are ad-hoc — no centralized shortcut registry

The keyboard shortcut `Ctrl+Enter` is now referenced in at least two places: the tooltip string in this component and the actual `keydown` event handler that triggers `handleRun`. If the shortcut is ever changed (e.g., to `Ctrl+Shift+Enter`), both locations must be updated manually. There is no single source of truth tying the displayed shortcut hint to the registered handler.

This is a broader architectural observation about the codebase, not just this PR, but this PR is a good opportunity to address it.

**Suggestion:**
```
Consider introducing a shortcut manifest:
```tsx
// constants/shortcuts.ts
export const SHORTCUTS = {
  RUN_PIPELINE: { key: 'Enter', ctrlKey: true },
} as const;

export function formatShortcut(shortcut: typeof SHORTCUTS[keyof typeof SHORTCUTS], isMac: boolean) {
  const mod = isMac ? '⌘' : 'Ctrl+';
  return `${mod}${shortcut.key}`;
}
```
Then both the tooltip and the event listener import from `SHORTCUTS`, ensuring they stay in sync.
```

File: node_modules/.package-lock.json [hunk 3]
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "imagelab",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}
```
ivantha: **[CRITICAL]** node_modules committed to version control

`node_modules/` should **never** be tracked in a git repository. Committing it causes:

- **Repository bloat**: Real `node_modules` trees can be hundreds of megabytes and contain thousands of platform-specific binaries.
- **Security risk**: Committing installed packages can accidentally lock in a vulnerable snapshot that bypasses `npm audit` or Dependabot workflows since the files are already in-tree.
- **Merge conflicts**: Lockfile and package metadata files inside `node_modules` will create constant noise in PRs and `git blame`.
- **CI/CD breakage**: Many pipelines clean the working tree with `git checkout .` or `git clean -fd`, which can delete or corrupt the installed packages if they are tracked.

The file committed here (`node_modules/.package-lock.json`) is npm's internal integrity manifest. Even though it is currently empty (`"packages": {}`), its presence in the diff proves that `node_modules/` is not excluded from git, and future `npm install` runs will silently populate and track the full dependency tree.

**Suggestion:**
```
1. Add `node_modules/` to `.gitignore` immediately:
```
# .gitignore
node_modules/
```
2. Remove the tracked file and purge it from git's index:
```bash
git rm -r --cached node_modules/
git commit -m "chore: remove node_modules from version control"
```
3. Verify with `git check-ignore -v node_modules/` that the directory is now properly ignored before merging this PR.
```

Comment (ivantha) [inline node_modules/.package-lock.json]: **[HIGH]** Missing or misconfigured .gitignore for node_modules

The presence of `node_modules/.package-lock.json` in this diff is a symptom of a missing root-level `.gitignore` rule. This is not an isolated accident — any developer who runs `npm install` in this repository will now have their entire `node_modules/` tree as untracked (or worse, tracked) files visible in `git status`. This creates a persistent friction point and a repeated opportunity for someone to accidentally commit real packages, including ones with known CVEs.

**Suggestion:**
```
Audit the `.gitignore` at the repository root and in `imagelab-frontend/`. At minimum it should contain:
```gitignore
# Dependencies
node_modules/
.pnp
.pnp.js

# Build outputs
dist/
build/

# Environment files
.env
.env.local
.env.*.local
```
Using `npx gitignore node` or the GitHub-provided Node.js `.gitignore` template is a quick way to get a comprehensive starting point.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

The core change — adding a keyboard shortcut hint to the Run Pipeline tooltip — is a small, well-motivated UX improvement that correctly aligns the button with the rest of the toolbar. However, the PR inadvertently commits `node_modules/.package-lock.json`, which strongly indicates that `node_modules/` is missing from `.gitignore`; this is a significant repository hygiene and latent security issue that must be remediated before merge. The functional change itself is low-risk but lacks automated test coverage and has a minor accessibility gap.

File: imagelab-frontend/src/components/Toolbar.tsx (hunk 2 above)
Comment (ivantha) [inline imagelab-frontend/src/components/Toolbar.tsx]: **[LOW]** No automated test coverage for tooltip content

The PR checklist explicitly marks "New tests added" as unchecked. While the change is simple, the tooltip text is user-facing and platform-conditional. Without a test, a future refactor that changes `mod` (e.g., switching from `'Ctrl+'` to `'Ctrl'`) could silently break the tooltip string without any CI signal.

This is especially relevant given that the entire motivation of this PR is consistency across toolbar buttons — a regression test would lock in that consistency guarantee.

**Suggestion:**
```
Add a unit test using your existing test framework (e.g., Vitest + React Testing Library):
```tsx
// Toolbar.test.tsx
import { render, screen } from '@testing-library/react';
import Toolbar from './Toolbar';

describe('Toolbar Run Pipeline button', () => {
  it('includes a keyboard shortcut hint in the tooltip', () => {
    render(<Toolbar workspace={mockWorkspace} />);
    const runBtn = screen.getByRole('button', { name: /run/i });
    expect(runBtn.title).toMatch(/Run Pipeline \((.+)Enter\)/);
  });

  it('shows Ctrl+ on non-Mac platforms', () => {
    // mock navigator.platform to 'Win32'
    // assert title contains 'Ctrl+Enter'
  });

  it('shows ⌘ on macOS', () => {
    // mock navigator.platform to 'MacIntel'
    // assert title contains '⌘Enter'
  });
});
```
```

Comment (ivantha) [inline imagelab-frontend/src/components/Toolbar.tsx]: **[LOW]** Platform detection via navigator.platform is deprecated

`navigator.platform` has been deprecated in the Web Platform specification (MDN marks it as deprecated since 2022). The recommended replacement is `navigator.userAgentData.platform` (with a fallback) or checking `navigator.userAgent` as a fallback. Using the deprecated API is not a blocker today but could cause linting warnings with strict browser-compat lint rules and will eventually break in future browser versions.

**Suggestion:**
```
Use the modern API with a fallback:
```tsx
const isMac = (
  (navigator as any).userAgentData?.platform ||
  navigator.platform
).toLowerCase().startsWith('mac');
```
Or use a well-maintained utility like `platform-detect` to avoid handling the branching logic manually.
```

Comment (ivantha) [inline imagelab-frontend/src/components/Toolbar.tsx]: **[INFO]** Accessibility: title attribute has limited screen reader support

The `title` attribute on interactive elements (`<button>`) is [not reliably announced by all screen readers](https://developer.paciellogroup.com/blog/2013/01/the-problem-with-the-title-attribute/), particularly on touch/mobile. This is a pre-existing issue across the entire toolbar (not introduced by this PR), but since this PR is improving discoverability of keyboard shortcuts, it is worth flagging.

For screen reader users who rely on keyboard navigation, not being able to hear the shortcut hint is an ironic accessibility gap.

**Suggestion:**
```
Consider replacing or supplementing the `title` attribute with an `aria-label` that includes the shortcut:
```tsx
<button
  aria-label={`Run Pipeline (${mod}Enter)`}
  title={`Run Pipeline (${mod}Enter)`}
  ...
>
```
For richer tooltip behavior with proper ARIA roles (e.g., `role="tooltip"` + `aria-describedby`), consider a small accessible tooltip component rather than relying solely on the native `title` attribute. This should be addressed as a follow-up for the entire toolbar.
```

Anirudh-x: Good point on the architectural concern. I've addressed this by extracting the platform detection logic into a shared useModKey hook (src/hooks/useModKey.ts), which is now the single source of truth for the modifier key string used across all tooltips and aria-labels. A full shortcut manifest can be a follow-up if the shortcut list grows.

ivantha:
CI is currently red (failing check: Frontend Lint). Could you take a look and push a fix? Happy to merge once CI is green.

ivantha:
Thanks @Anirudh-x! The keyboard shortcut hint for Run Pipeline is a small UI improvement, but this PR conflicts with main and has multiple CI failures. The base feature (keyboard shortcuts) was merged as #74. Closing this — feel free to rebase and open a small focused PR with just the tooltip change!

Issue #39, opened 2026-02-25
GSoC 2026 — ImageLab: Interactive Learning Mode


ivantha:
## Title

Interactive Learning Mode: Per-Step Visual Preview, Histogram Analysis, Pipeline Persistence, Batch Processing, and Custom Composite Operators

## Abstract

ImageLab is a block-based image processing tool built on Blockly and OpenCV. Students drag operators into a pipeline and run it against an uploaded image. Today the pipeline is fire-and-forget: you press **Run**, the backend executes every step, and you see only the final result. There is no way to inspect what a single Gaussian blur or threshold does in isolation, no quantitative analysis of pixel values, and no ability to save or share a pipeline.

This project transforms ImageLab into a genuine **learning environment** by adding five capabilities:

1. **Per-step intermediate previews** — the backend returns the image after every operator, and the frontend renders a scrollable, step-by-step visualization.
2. **Image analysis panel** — a collapsible panel showing a per-channel RGB histogram (rendered via Canvas), image metadata (dimensions, channels, dtype, min/max/mean pixel values) for any selected step.
3. **Pipeline persistence** — SQLModel database models, CRUD REST endpoints, Blockly workspace serialization, and a save/load/share-via-URL workflow.
4. **Batch processing engine** — apply a saved pipeline to multiple images at once, with concurrent execution, progress tracking, and ZIP download of results.
5. **Custom composite operators (macros)** — select a sub-chain of blocks, name it, and save it as a reusable "macro" that appears in the Blockly sidebar as a custom block.

## Goals

- Let students see the effect of every single operator, not just the final output
- Provide quantitative tools (histogram, statistics) for understanding image transformations
- Allow saving, loading, and sharing named pipelines so students can revisit and collaborate
- Enable batch processing of multiple images through a saved pipeline for real-world workflows
- Let users create reusable composite operators (macros) to build higher-level abstractions

## Expected Outcomes

- Students can inspect every operator's effect step-by-step instead of seeing only the final image
- Quantitative analysis (histogram, statistics) is available for every intermediate step
- Pipelines can be saved, loaded, and shared via URL, enabling collaboration and teaching workflows
- ImageLab's first database usage sets the foundation for future features (user accounts, pipeline ratings)
- Batch processing allows applying saved pipelines to multiple images concurrently with progress tracking and ZIP download
- Custom composite operators (macros) teach abstraction by letting users package sub-chains as reusable blocks in the Blockly sidebar

## Prerequisites for Applicants

[...]

PR #98, opened 2026-02-28
feat: Interactive Learning Mode — per-step previews and image analysis panel (#39)

Anirudh-x:
## Description

Implements Feature 1 (per-step intermediate previews) and Feature 2 (image analysis panel with histogram) from the Interactive Learning Mode project.

Previously the pipeline was fire-and-forget — users only saw the final result. This change lets users toggle **Step Preview** in the toolbar, run the pipeline, and inspect the output of every single operator with a thumbnail, pixel statistics, and a per-channel BGR/grayscale histogram.

**Backend:**
- Extended `PipelineRequest` with `include_intermediates` flag
- Added `ImageStats` and `StepResult` Pydantic models to `PipelineResponse`
- Added `compute_image_stats()` utility to compute dimensions, channels, dtype, min/max/mean, and 256-bucket per-channel histograms using OpenCV
- `pipeline_executor` now captures each operator's output when the flag is enabled

**Frontend:**
- Extended `PipelineRequest` / `PipelineResponse` types
- Added `intermediates`, `showStepPreviews`, `selectedStepIndex` to the Zustand store
- New `HistogramCanvas` component — draws BGR/grayscale histograms on a `<canvas>`
- New `StepPreviewList` component — scrollable panel with per-step thumbnails; click to expand stats + histogram
- Toolbar: added **Step Preview** toggle button; passes `include_intermediates` on run
- Layout: renders `StepPreviewList` as a collapsible side panel when the toggle is active

Fixes #39

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

Manual testing steps performed:
1. Ran `uv run uvicorn app.main:app --reload --port 4100` and `npm run dev`
2. Loaded an image via the Read Image block, built a multi-step pipeline (blur → grayscale → threshold)
3. Toggled **Step Preview** on in the toolbar and clicked **Run**
4. Verified the step panel appeared with correct thumbnails for each operator
5. Clicked each card to expand — confirmed dimensions, dtype, min/max/mean, and histogram were accurate
6. Toggled Step Preview off — confirmed panel disappears and pipeline still runs normally
7. Ran `npm run build` — zero TypeScript errors
8. Ran `uv run python -c "from app.models.pipeline import ...; from app.utils.image import compute_image_stats; ..."` — all imports resolved

## Checklist

- [x] Tests pass locally


imagelab-frontend/src/components/Toolbar.tsx
```diff
@@ -118,6 +128,21 @@ export default function Toolbar({ workspace }: ToolbarProps) {
...
+        onClick={() => setShowStepPreviews(!showStepPreviews)}
+        className={`flex items-center gap-1 px-2 py-1 rounded text-xs font-medium border transition-colors ${showStepPreviews
+          ? "bg-indigo-50 border-indigo-300 text-indigo-600"
+          : "bg-white border-gray-300 text-gray-500 hover:bg-gray-50"
+          }`}
+        title="Toggle per-step intermediate previews"
+      >
+        <Layers size={13} />
+        Step Preview
+      </button>
```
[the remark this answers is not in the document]
Anirudh-x: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/c2siorg/imagelab/pull/98#discussion_r2867170114)

imagelab-frontend/src/components/Preview/StepPreviewList.tsx
```diff
@@ -0,0 +1,143 @@
...
+ *  a simpler display label by dropping the category prefix (e.g. "applygaussianblur"). */
+function operatorLabel(type: string): string {
+  const underscoreIndex = type.indexOf("_");
+  if (underscoreIndex === -1) {
+    return type;
+  }
+  const name = type.slice(underscoreIndex + 1).trim();
+  return name || type;
+}
+}
```
ivantha: **[CRITICAL]** Dangling closing brace causes TypeScript parse error

After the closing `}` of `operatorLabel` (line 15) there is a second, orphaned `}` on line 16 with no matching opening brace. This is a top-level syntax error that TypeScript cannot parse — the module cannot be compiled.

The diff shows two consecutive `+}` lines:
```
  return name || type;
}   // ← correct: closes operatorLabel
}   // ← stray: no matching {
```
This contradicts the PR author's claim of "zero TypeScript errors". The most likely explanation is that this was introduced by a merge/diff artifact, but as written the file is unparseable.

**Suggestion:**
```
Remove the extra `}` so that the module-level scope is correct:
```typescript
function operatorLabel(type: string): string {
  const underscoreIndex = type.indexOf("_");
  if (underscoreIndex === -1) {
    return type;
  }
  const name = type.slice(underscoreIndex + 1).trim();
  return name || type;
}
// No extra }

/** One card in the step list. */
function StepCard(...){
```
```

File: imagelab-backend/app/utils/image.py [hunk 3]
```diff
@@ -18,3 +20,43 @@ def encode_image_base64(image: np.ndarray, fmt: str = "png") -> str:
...
+        # Grayscale
+        height, width = image.shape
+        channels = 1
+        safe_img = np.clip(image, 0, 255).astype(np.uint8)
+        hist = cv2.calcHist([safe_img], [0], None, [256], [0, 256])
+        histograms = [hist[:, 0].astype(int).tolist()]
+    else:
+        height, width = image.shape[:2]
+        channels = image.shape[2]
+        safe_img = np.clip(image, 0, 255).astype(np.uint8)
```
ivantha: **[HIGH]** Float images with values in [0, 1] produce degenerate histograms

The histogram computation blindly clips the array to [0, 255] and casts to `uint8` before calling `cv2.calcHist`:
```python
safe_img = np.clip(image, 0, 255).astype(np.uint8)
hist = cv2.calcHist([safe_img], [0], None, [256], [0, 256])
```
For a normalized `float32` image whose pixel values are in `[0.0, 1.0]` — common output from operators like `cv2.normalize` — `np.clip` preserves the values unchanged, and `.astype(np.uint8)` then truncates every pixel to either **0** or **1**. The resulting histogram has ~100% of pixels piled into bins 0–1 and zero counts everywhere else, which is completely misleading.

Noteworthy inconsistency: the `min`/`max`/`mean` statistics are computed from the **original** `image` array (correct), while the histograms are computed from the clipped `safe_img` (incorrect for floats). The two numbers reported to the user therefore describe different data.

**Suggestion:**
```
Detect the image dtype and scale appropriately before computing histograms. For float images, determine the actual value range and either normalize to [0, 255] or use `cv2.calcHist` with the correct range:
```python
def _to_uint8_for_hist(image: np.ndarray) -> np.ndarray:
    """Scale image to uint8 for histogram computation, handling both uint8 and float dtypes."""
    if image.dtype == np.uint8:
        return image
    img_min, img_max = float(image.min()), float(image.max())
    if img_max == img_min:
        return np.zeros_like(image, dtype=np.uint8)
    scaled = (image - img_min) / (img_max - img_min) * 255.0
    return np.clip(scaled, 0, 255).astype(np.uint8)
```
Then use `_to_uint8_for_hist(image)` instead of the current `np.clip(image, 0, 255).astype(np.uint8)`. Document that histograms for float images represent normalised pixel intensity, not raw values.
```

File: imagelab-backend/app/utils/image.py [hunk 4]
```diff
@@ -3,6 +3,8 @@
 import cv2
 import numpy as np
 
+from app.models.pipeline import ImageStats
+
```
ivantha: **[MEDIUM]** Utility module imports from models — architectural inversion

`app/utils/image.py` now imports `ImageStats` from `app/models/pipeline.py`:
```python
from app.models.pipeline import ImageStats
```
This creates an upward dependency: a low-level utility module depends on a domain-specific model defined in a higher-level layer. This is an architectural inversion and makes the `image` utility module less reusable (it is now coupled to the pipeline domain). It also increases the risk of circular imports if `models/pipeline.py` ever needs to call image utilities.

The conventional FastAPI project layout has `utils` as a leaf node that other layers can safely import without pulling in the whole application model graph.

**Suggestion:**
```
Either:
1. **Return a plain dict/dataclass from `compute_image_stats`** and have the caller (e.g., `pipeline_executor.py`) construct the `ImageStats` Pydantic model, OR
2. **Move `ImageStats` to a separate `app/models/image.py`** or `app/schemas/image.py` module that can be imported by both utils and pipeline models without creating a dependency cycle.
```

ivantha (commented):
## ◎ Argus Review

This PR introduces a well-structured interactive learning mode that exposes per-step pipeline intermediates with thumbnails, pixel statistics, and histograms. The core concept is solid — the Pydantic models are clean, the Zustand store is updated comprehensively with correct cleanup in `setOriginalImage`/`clearImage`/`resetAll`, and the `HistogramCanvas` component correctly handles device-pixel-ratio scaling and channel ordering. However, there is a **syntax error** (dangling `}`) in `StepPreviewList.tsx` that would prevent TypeScript compilation, a **significant correctness bug** in `compute_image_stats` where float images with values outside [0, 255] produce meaningless histograms, and several medium-priority issues around BGRA channel rendering, payload size limits, missing `type="button"` on the toggle, and an architectural inversion between `utils` and `models`. These should be resolved before merging.

imagelab-backend/app/services/pipeline_executor.py
```diff
@@ -33,6 +37,26 @@ def execute_pipeline(request: PipelineRequest) -> PipelineResponse:
...
+            except Exception as capture_err:
+                # Non-fatal: log and skip this step's intermediate instead of silently dropping it
+                logging.getLogger(__name__).warning(
+                    "Failed to capture intermediate for step %d (%s): %s: %s",
+                    i, step.type, type(capture_err).__name__, capture_err
+                )
+
     try:
         encoded = encode_image_base64(image, request.image_format)
     except Exception as e:
```
ivantha: **[MEDIUM]** No limit on intermediate payload — can produce very large responses

When `include_intermediates=True`, every non-NOOP step triggers both a full `cv2.imencode` (O(W×H)) and a multi-channel `cv2.calcHist` pass, with the base64-encoded image appended to the response. For a long pipeline (e.g., 20 steps) on a high-resolution image (e.g., 4096×4096), this could:
- Encode and hold many large base64 strings in memory simultaneously before sending the response
- Produce an HTTP response body of hundreds of MB
- Block the event loop for several seconds since FastAPI with uvicorn uses async but the encoding itself is synchronous/CPU-bound

There is currently no guard against any of these scenarios.

**Suggestion:**
```
Consider one or more of the following mitigations:
1. **Add a max-steps guard**: reject or warn if `len(request.pipeline) > N` when `include_intermediates=True`.
2. **Downscale thumbnails** in the backend before encoding intermediates (e.g., `cv2.resize` to max 256×256) — the thumbnails are shown at `max-h-24` anyway.
3. **Stream the intermediates** as Server-Sent Events or a WebSocket stream rather than collecting all of them in one giant JSON response.
4. **Limit total payload** (e.g., skip encoding intermediates if the estimated size exceeds a threshold).
```

imagelab-frontend/src/store/pipelineStore.ts
```diff
@@ -61,7 +83,7 @@ export const usePipelineStore = create<PipelineState>((set) => ({
   clearImage: () => {
     const state = usePipelineStore.getState();
     if (state._imageResetFn) state._imageResetFn();
```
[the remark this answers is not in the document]
Anirudh-x: Good suggestion on the INITIAL_STATE constant — will keep that in mind as the store grows to avoid accidentally omitting a field from one of the reset paths.

imagelab-frontend/src/components/Layout.tsx
```diff
@@ -27,6 +30,19 @@ export default function Layout() {
               <div ref={containerRef} className="flex-1" />
               <InfoPane />
             </div>
+            {/* Step-by-step preview panel (shown when toggle is on) */}
+            {showStepPreviews && (
+              <div className="w-64 flex-shrink-0 h-full bg-white border-l border-gray-200 flex flex-col">
```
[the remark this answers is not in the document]
Anirudh-x: Fixed. Replaced the hardcoded w-64 with a draggable resize handle on the left edge of the panel. Width is clamped between 180 px and 480 px (default 256 px). The mousemove/mouseup listeners are attached to window so the drag remains smooth even if the cursor moves quickly off the handle.

ivantha (commented):
## ◎ Argus Review

This PR delivers a well-architected Interactive Learning Mode with good patterns throughout — opt-in flag design, non-fatal intermediate capture with structured logging, and the `INITIAL_STATE`/`IMAGE_RESET` store refactor are all commendable. However, there is one **critical blocker**: `StepResult` and `ImageStats` are used in `pipeline_executor.py` but never imported, meaning every request with `include_intermediates=True` raises a `NameError` and the feature is completely non-functional. Two additional high-severity frontend issues exist — window drag-event listeners are not cleaned up on unmount (memory leak) and the `useCallback` dependency on frequently-changing state causes excessive recreation during resize. Beyond these, the PR would benefit from at minimum minimal unit tests for the new backend utilities and returning partial intermediates on pipeline errors.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes detected — all API changes are additive with safe defaults: **Previous behavior:** `PipelineRequest` had no `include_intermediates` field; `PipelineResponse` had no `intermediates` field.

**New behavior:** Both fields are added as optional with defaults (`False` / `None`).

**Who is affected:** All consumers of the HTTP API and the Zustand store. Because both additions use safe defaults, existing clients receive responses identical to pre-PR behaviour.

**Migration path:** None required. Clients opt in by sending `include_intermediates: true`.


### PR Description Issues

- **[INFO]** No tests added — PR acknowledges this but provides no follow-up plan: The "New tests added" checkbox is unchecked. The PR introduces ~150 lines of new backend logic (`compute_image_stats`, `_thumbnail`, intermediate capture) with zero test coverage. The description does not reference a follow-up issue or milestone to address this gap.
  > **Suggestion:** Either add minimal unit tests before merging, or open a follow-up issue and reference it:

> **Follow-up:** Unit tests for `compute_image_stats` and `_thumbnail` tracked in #NN.
- **[INFO]** `.gitignore` and `vitest` constraint changes are undocumented in the PR description: The diff includes two unannounced changes:
1. Adding `node_modules/` to `.gitignore` — implies it was missing, which is unusual.
2. Tightening `vitest` from `^4.0.17` (minor+patch) to `~4.0.17` (patch-only) in `package-lock.json`.

Neither is mentioned in the PR body, which may surprise reviewers and future contributors.
  > **Suggestion:** Add a brief housekeeping note to the description:

> **Housekeeping:**
> - Added `node_modules/` to `.gitignore` (was previously missing)
> - Pinned `vitest` to `~4.0.17` (patch updates only) to reduce unexpected churn

ivantha:
Holding this PR — it is linked to a GSoC-labeled issue (#39) and should go through GSoC evaluation rather than maintainer merge.

ivantha:
Thanks @Anirudh-x! Closing this as it has multiple CI failures (frontend lint, backend lint, rebase check) and the per-step preview feature is now addressed by cleaner PRs (#233, #269). If you'd like to continue, please rebase on main, fix the lint errors, and open a fresh PR — we'd love to review the histogram panel concept!

PR #110, opened 2026-03-01
Fix Sobel and Scharr operators returning raw float64 output

anusha19murthy:
Convert derivative results to uint8 via np.absolute() before returning, matching the existing Laplacian operator pattern.

## Description

Sobel and Scharr derivative operators were returning raw float64 arrays with negative values, causing silent image corruption or encoding failures. This fix converts results to uint8 via np.absolute() before returning, matching the existing Laplacian operator pattern.
**Root Cause**: cv2.Sobel() and cv2.Scharr() with ddepth=cv2.CV_64F produce signed float64 output. When passed to cv2.imencode(), negative values were clipped to 0 (destroying half the gradient data) or causing encode failures.

---

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

- [X] Existing tests pass
- [ ] New tests added
- [X] Manual testing

---

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review
- [X] My changes generate no new warnings
- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py
+++ b/imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py
@@ -15,9 +15,6 @@ class ScharrDerivative(BaseOperator):
         if ddepth == 0:
             ddepth = cv2.CV_64F
 
-        if direction == "HORIZONTAL":
-            result = cv2.Scharr(image, ddepth, 1, 0)
-        else:
-            result = cv2.Scharr(image, ddepth, 0, 1)
+        result = cv2.Scharr(image, ddepth, 1, 0) if direction == "HORIZONTAL" else cv2.Scharr(image, ddepth, 0, 1)
 
         return cv2.convertScaleAbs(result)


imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py
```diff
@@ -12,10 +12,12 @@ def compute(self, image: np.ndarray) -> np.ndarray:
...
-            return cv2.Sobel(image, ddepth, 1, 0)
+            result = cv2.Sobel(image, ddepth, 1, 0)
         elif direction == "VERTICAL":
-            return cv2.Sobel(image, ddepth, 0, 1)
+            result = cv2.Sobel(image, ddepth, 0, 1)
         else:
             sobel_x = cv2.Sobel(image, ddepth, 1, 0)
             sobel_y = cv2.Sobel(image, ddepth, 0, 1)
-            return cv2.addWeighted(sobel_x, 0.5, sobel_y, 0.5, 0)
+            result = cv2.addWeighted(sobel_x, 0.5, sobel_y, 0.5, 0)
```
ivantha: **[MEDIUM]** COMBINED direction uses incorrect gradient magnitude formula — diagonal edges cancel out

The `COMBINED` (else) branch computes:
```python
result = cv2.addWeighted(sobel_x, 0.5, sobel_y, 0.5, 0)
# i.e. result = 0.5*Gx + 0.5*Gy
```
This is **not** a gradient magnitude. Because `Gx` and `Gy` retain their signs, they can cancel each other. For example:
- `Gx = 200` (strong right-facing gradient), `Gy = -200` (strong downward gradient)
- `addWeighted` result = **0** → after `np.absolute()` → **0**
- Correct magnitude: `sqrt(200² + 200²) ≈ 283`

Diagonal edges would appear **invisible** in COMBINED mode. This is a pre-existing bug, but this PR wraps it in a new conversion layer without correcting it — and the bug now surfaces more visibly since the output reaches users as uint8 images.

The standard formulas are:
```python
# Euclidean (preferred)
result = np.sqrt(sobel_x**2 + sobel_y**2)
# Manhattan approximation
result = np.absolute(sobel_x) + np.absolute(sobel_y)
```

**Suggestion:**
```
Replace the `COMBINED` branch computation:
```python
# Before applying convertScaleAbs:
result = np.sqrt(sobel_x**2 + sobel_y**2)
# or
result = np.absolute(sobel_x) + np.absolute(sobel_y)
```
Then apply `cv2.convertScaleAbs(result)` as the return.
```

File: imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py [hunk 2]
```diff
@@ -12,10 +12,12 @@ def compute(self, image: np.ndarray) -> np.ndarray:
...
+
+        return np.uint8(np.absolute(result))
```
ivantha: **[MEDIUM]** No new tests added for the dtype conversion and overflow boundary conditions

The PR checklist explicitly acknowledges that no new tests were added. Given the nature of this bug — silent data corruption that depends on precise numeric thresholds — tests are essential to prevent regression. Missing test cases include:

1. **Return dtype assertion**: `assert result.dtype == np.uint8`
2. **Value range assertion**: `assert result.min() >= 0 and result.max() <= 255`
3. **Overflow boundary**: A synthetic image with a strong known gradient (e.g., black-to-white step) should produce values ≤ 255, not wrapped artifacts.
4. **All directions**: HORIZONTAL, VERTICAL, and COMBINED each tested.
5. **Both color and grayscale inputs** (since `ddepth` is conditionally set based on image channels).

**Suggestion:**
```
Add unit tests such as:
```python
def test_sobel_returns_uint8_in_range():
    # High-contrast step image produces strong gradients
    img = np.zeros((50, 50, 3), dtype=np.uint8)
    img[:, 25:] = 255
    op = SobelDerivative(params={"direction": "HORIZONTAL", "order": 1})
    result = op.compute(img)
    assert result.min() >= 0
    assert result.max() <= 255
```
```

Comment (ivantha) [inline imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py]: **[MEDIUM]** Information loss: absolute value discards sign; normalization not considered for visualization

Taking `np.absolute()` discards gradient direction (sign). This is acceptable for visualization, but the clamp-to-uint8 step (whether via `np.uint8()` or `np.clip()`) further discards relative gradient magnitude by hard-clipping at 255. For example, a gradient of 800 and a gradient of 300 both clip to 255, making them visually indistinguishable.

For visualization, **normalization** to [0, 255] is often preferred because it preserves relative edge strengths across the full display range:
```python
abs_result = np.absolute(result)
normalized = cv2.normalize(abs_result, None, 0, 255, cv2.NORM_MINMAX)
return np.uint8(normalized)
```
This is a design decision, not strictly a bug, but worth raising with the author.

**Suggestion:**
```
If this output is intended for visualization (image display/encoding), consider normalizing rather than clipping:
```python
abs_result = np.absolute(result)
return np.uint8(cv2.normalize(abs_result, None, 0, 255, cv2.NORM_MINMAX))
```
Documentation should clarify the intended semantics of the output.
```

File: imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py [hunk 3]
```diff
@@ -12,6 +12,8 @@ def compute(self, image: np.ndarray) -> np.ndarray:
             ddepth = cv2.CV_64F
 
         if direction == "HORIZONTAL":
-            return cv2.Scharr(image, ddepth, 1, 0)
+            result = cv2.Scharr(image, ddepth, 1, 0)
         else:
-            return cv2.Scharr(image, ddepth, 0, 1)
+            result = cv2.Scharr(image, ddepth, 0, 1)
+
+        return np.uint8(np.absolute(result))
```
ivantha: **[LOW]** Return type semantics changed without updating docstrings or type annotations

The function previously returned a `float64` array with signed values. It now returns a `uint8` array with absolute values. If the function has a docstring or type annotation (not visible in the diff), these should be updated to reflect:
- Return dtype: `np.ndarray` with `dtype=np.uint8`
- Value range: `[0, 255]`
- Semantic note: values represent gradient *magnitude* (sign discarded)

Callers relying on the old float64 output for further computation (not just display) could silently receive incorrect results after this change.

**Suggestion:**
```
Add or update the function docstring:
```python
    """
    Compute Scharr derivative.
    Returns a uint8 ndarray in range [0, 255] representing gradient magnitude.
    Sign information is discarded; intended for visualization.
    """
```
```

File: imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py (hunk 1 above)
Comment (ivantha) [inline imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py]: **[LOW]** COMBINED path computes intermediate float64 gradients redundantly when absolute is taken anyway

In the COMBINED branch:
```python
sobel_x = cv2.Sobel(image, ddepth, 1, 0)  # float64
sobel_y = cv2.Sobel(image, ddepth, 0, 1)  # float64
result = cv2.addWeighted(sobel_x, 0.5, sobel_y, 0.5, 0)
```
If the fix moved to the correct `np.sqrt(sobel_x**2 + sobel_y**2)` formula, the intermediate `result` variable could be inlined. Also, `ddepth` is only assigned conditionally in one branch but used in all — confirm there are no code paths where `ddepth` could be uninitialized (not apparent from the diff, but worth checking the full function).

**Suggestion:**
```
After correcting the formula, simplify to:
```python
result = np.hypot(
    cv2.Sobel(image, ddepth, 1, 0),
    cv2.Sobel(image, ddepth, 0, 1)
)
return cv2.convertScaleAbs(result)
```
```

File: imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py (hunk 3 above)
Comment (ivantha) [inline imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py]: **[INFO]** Positive: Refactoring from early returns to single return point improves readability

Replacing multiple `return cv2.Scharr(...)` early-return statements with a single `return` at the end of the function is a good pattern. It makes it easy to see at a glance that all branches produce a consistently-transformed result, and it avoids the risk of future developers adding a new branch and forgetting to apply the conversion. This structure also makes it trivial to add logging, metrics, or post-processing in one place later.

**Suggestion:**
```
No change needed — this is a positive pattern. Continue using it across similar operators.
```

PR #112, opened 2026-03-01
feature: add pipeline share code feature

### LINKED ISSUE #111: Feature - Add Share Pipeline feature

Currently, there is no way to share a pipeline with others. If someone builds a pipeline and wants others to replicate it exactly, they would have to manually recreate it block by block.

A Share Pipeline feature accessible from the toolbar. Clicking the share icon generates an encoded code representing the current workspace. Anyone can paste that code into the Load from Code input to instantly recreate the exact same pipeline in their own workspace. The image filename is intentionally excluded from the code so each user loads their own image.

Saving pipelines to the backend database with short codes was considered, but encoding the pipeline directly into the code string was preferred as it requires no storage, works permanently, and has no backend dependencies.

**Additional context**

This is useful for anyone who wants to share their pipeline with others, for example, a teacher demonstrating a specific image processing pipeline to students, or a contributor sharing a reproducible example when reporting an issue.

I will be working on this issue.

xxivani:
## Description
Adds a Share Pipeline feature to the toolbar, allowing users to generate an encoded code representing their current Blockly workspace. Anyone with the code can paste it into the Load from Code input to instantly recreate the exact same pipeline in their own workspace. The image filename is excluded from the code so each user loads their own image.

Fixes #111 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Generated a code from a pipeline with multiple blocks and verified it decodes correctly into the same workspace.
Verified that the image filename is cleared after loading, showing "No image" instead.
Verified that pasting an invalid code shows an appropriate error message.
Verified that attempting to generate a code with an empty workspace shows a prompt to add blocks first.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 

https://github.com/user-attachments/assets/cce8d5e6-49b5-4cd5-9a40-7d32843b198d

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-frontend/src/components/SharePipelineModal.tsx
+++ b/imagelab-frontend/src/components/SharePipelineModal.tsx
@@ -1,6 +1,9 @@
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import * as Blockly from "blockly";
 import { X, Copy, Check, Share2, Upload } from "lucide-react";
+
+const READ_IMAGE_BLOCK_TYPE = "basic_readimage";
+const FILENAME_LABEL_FIELD = "filename_label";
 
 interface SharePipelineModalProps {
   workspace: Blockly.WorkspaceSvg | null;
@@ -16,7 +19,7 @@ function compressToCode(workspace: Blockly.WorkspaceSvg): string {
   return btoa(binary);
 }
 
-function decompressFromCode(code: string): object {
+function decompressFromCode(code: string): unknown {
   const binary = atob(code.trim());
   const bytes = new Uint8Array(binary.length);
   for (let i = 0; i < binary.length; i++) {
@@ -30,8 +33,18 @@ export default function SharePipelineModal({ workspace, onClose }: SharePipeline
   const [generatedCode, setGeneratedCode] = useState<string | null>(null);
   const [inputCode, setInputCode] = useState("");
   const [copied, setCopied] = useState(false);
+  const [copyError, setCopyError] = useState<string | null>(null);
   const [loadError, setLoadError] = useState<string | null>(null);
   const [loadSuccess, setLoadSuccess] = useState(false);
+
+  // Close on Escape key
+  useEffect(() => {
+    const handler = (e: KeyboardEvent) => {
+      if (e.key === "Escape") onClose();
+    };
+    document.addEventListener("keydown", handler);
+    return () => document.removeEventListener("keydown", handler);
+  }, [onClose]);
 
   const handleGenerate = () => {
     if (!workspace) return;
@@ -46,24 +59,49 @@ export default function SharePipelineModal({ workspace, onClose }: SharePipeline
 
   const handleCopy = async () => {
     if (!generatedCode) return;
-    await navigator.clipboard.writeText(generatedCode);
-    setCopied(true);
-    setTimeout(() => setCopied(false), 2000);
+    try {
+      await navigator.clipboard.writeText(generatedCode);
+      setCopied(true);
+      setCopyError(null);
+      setTimeout(() => setCopied(false), 2000);
+    } catch {
+      setCopyError("Could not copy to clipboard. Please select and copy manually.");
+    }
   };
 
   const handleLoad = () => {
     if (!workspace || !inputCode.trim()) return;
     setLoadError(null);
     setLoadSuccess(false);
+
+    // Confirm before overwriting existing workspace
+    const existingBlocks = workspace.getAllBlocks(false);
+    if (existingBlocks.length > 0) {
+      if (!window.confirm("Loading a pipeline will replace your current workspace. Continue?"))
+        return;
+    }
+
     try {
       const state = decompressFromCode(inputCode.trim());
-      workspace.clear();
-      Blockly.serialization.workspaces.load(state, workspace);
 
-      const readImageBlocks = workspace.getBlocksByType("basic_readimage", false);
-      readImageBlocks.forEach((block) => {
-        const field = block.getField("filename_label");
-        if (field) field.setValue("No image");
+      // Save snapshot before mutating so we can restore on failure
+      const snapshot = Blockly.serialization.workspaces.save(workspace);
+      workspace.clear();
+
+      try {
+        Blockly.serialization.workspaces.load(
+          state as Parameters<typeof Blockly.serialization.workspaces.load>[0],
+          workspace,
+        );
+      } catch (loadErr) {
+        // Restore original workspace on failure
+        Blockly.serialization.workspaces.load(snapshot, workspace);
+        throw loadErr;
+      }
+
+      // Clear filename from any read image blocks
+      workspace.getBlocksByType(READ_IMAGE_BLOCK_TYPE, false).forEach((block) => {
+        block.getField(FILENAME_LABEL_FIELD)?.setValue("No image");
       });
 
       setLoadSuccess(true);
@@ -74,8 +112,18 @@ export default function SharePipelineModal({ workspace, onClose }: SharePipeline
   };
 
   return (
-    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/40">
-      <div className="bg-white rounded-xl shadow-2xl w-full max-w-md mx-4 overflow-hidden">
+    <div
+      className="fixed inset-0 z-50 flex items-center justify-center bg-black/40"
+      onClick={onClose}
+      role="presentation"
+    >
+      <div
+        role="dialog"
+        aria-modal="true"
+        aria-label="Share Pipeline"
+        className="bg-white rounded-xl shadow-2xl w-full max-w-md mx-4 overflow-hidden"
+        onClick={(e) => e.stopPropagation()}
+      >
         {/* Header */}
         <div className="flex items-center justify-between px-5 py-4 border-b border-gray-200">
           <div className="flex items-center gap-2">
@@ -83,11 +131,13 @@ export default function SharePipelineModal({ workspace, onClose }: SharePipeline
             <h2 className="text-sm font-semibold text-gray-800">Share Pipeline</h2>
           </div>
           <button
+            type="button"
             onClick={onClose}
             className="p-1 rounded hover:bg-gray-100 text-gray-400 hover:text-gray-600 transition-colors"
             title="Close"
+            aria-label="Close"
           >
-            <X size={16} />
+            <X size={16} aria-hidden="true" />
           </button>
         </div>
 
@@ -104,6 +154,7 @@ export default function SharePipelineModal({ workspace, onClose }: SharePipeline
             </div>
 
             <button
+              type="button"
               onClick={handleGenerate}
               disabled={!workspace}
               className="w-full py-2 px-3 rounded-lg text-sm font-medium text-white bg-indigo-500 hover:bg-indigo-600 disabled:opacity-50 disabled:cursor-not-allowed transition-colors"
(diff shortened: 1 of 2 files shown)


imagelab-frontend/src/components/SharePipelineModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+  workspace: Blockly.WorkspaceSvg | null;
+  onClose: () => void;
+}
+
+function compressToCode(workspace: Blockly.WorkspaceSvg): string {
+  const state = Blockly.serialization.workspaces.save(workspace);
+  const json = JSON.stringify(state);
+  const bytes = new TextEncoder().encode(json);
+  let binary = "";
+  bytes.forEach((b) => (binary += String.fromCharCode(b)));
```
ivantha: **[LOW]** O(n²) string concatenation in compressToCode

The pattern `bytes.forEach((b) => (binary += String.fromCharCode(b)))` builds a string by repeated concatenation inside a loop. In V8 and other JS engines, string concatenation in a tight loop may cause quadratic performance due to repeated re-allocation and copying. For large pipelines with many blocks and properties, this could noticeably stall the UI thread.

**Suggestion:**
```
Use `String.fromCharCode.apply` or `Array.from` with `join` to build the binary string in one allocation:
```typescript
function compressToCode(workspace: Blockly.WorkspaceSvg): string {
  const state = Blockly.serialization.workspaces.save(workspace);
  const json = JSON.stringify(state);
  const bytes = new TextEncoder().encode(json);
  // Single allocation instead of repeated concatenation
  const binary = String.fromCharCode(...bytes);
  return btoa(binary);
}
```
Note: For very large byte arrays, spread syntax may hit call stack limits. A safer alternative: `Array.from(bytes, b => String.fromCharCode(b)).join('')`.
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx [hunk 2]
```diff
@@ -0,0 +1,196 @@
...
+
+function decompressFromCode(code: string): object {
+  const binary = atob(code.trim());
+  const bytes = new Uint8Array(binary.length);
+  for (let i = 0; i < binary.length; i++) {
+    bytes[i] = binary.charCodeAt(i);
+  }
+  const json = new TextDecoder().decode(bytes);
+  return JSON.parse(json);
+}
```
ivantha: **[LOW]** Functions named 'compress/decompress' perform no actual compression

Both `compressToCode` and `decompressFromCode` only do base64 encoding/decoding — there is no compression step. The names imply gzip, zlib, LZ-string, or similar compression, which misleads future maintainers. For a large pipeline, the generated code will be very long (base64 overhead is ~33% larger than the raw JSON), making it harder to share via chat, email, or URL parameters. This is a naming correctness issue and a UX concern.

**Suggestion:**
```
Either rename the functions to reflect what they actually do (`encodePipelineToBase64` / `decodePipelineFromBase64`), or add real compression using a library like `lz-string` to keep the codes short:
```typescript
import LZString from 'lz-string';

function compressToCode(workspace: Blockly.WorkspaceSvg): string {
  const state = Blockly.serialization.workspaces.save(workspace);
  return LZString.compressToEncodedURIComponent(JSON.stringify(state));
}

function decompressFromCode(code: string): unknown {
  const json = LZString.decompressFromEncodedURIComponent(code.trim());
  if (!json) throw new Error("Invalid or corrupted code");
  return JSON.parse(json);
}
```
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx [hunk 3]
```diff
@@ -0,0 +1,196 @@
...
+    const code = compressToCode(workspace);
+    setGeneratedCode(code);
+  };
+
+  const handleCopy = async () => {
+    if (!generatedCode) return;
+    await navigator.clipboard.writeText(generatedCode);
+    setCopied(true);
+    setTimeout(() => setCopied(false), 2000);
+  };
```
ivantha: **[LOW]** setTimeout callback can fire after component unmounts

Two `setTimeout` calls — `setCopied(false)` on line 51 and `onClose()` on line 70 — are created without cleanup. If the modal is unmounted before the timeout fires (e.g., the parent closes it through another code path), the callback for `setCopied` may attempt a state update on an unmounted component. React 18 no longer warns about this specific case, but the `onClose()` callback could also execute unexpectedly, triggering effects in the parent.

**Suggestion:**
```
Use `useEffect` cleanup for timers:
```typescript
const copyTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);

const handleCopy = async () => {
  if (!generatedCode) return;
  try {
    await navigator.clipboard.writeText(generatedCode);
    setCopied(true);
    if (copyTimeoutRef.current) clearTimeout(copyTimeoutRef.current);
    copyTimeoutRef.current = setTimeout(() => setCopied(false), 2000);
  } catch { /* ... */ }
};

useEffect(() => {
  return () => {
    if (copyTimeoutRef.current) clearTimeout(copyTimeoutRef.current);
  };
}, []);
```
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx [hunk 4]
```diff
@@ -0,0 +1,196 @@
...
+
+      setLoadSuccess(true);
+      setTimeout(() => onClose(), 800);
+    } catch {
+      setLoadError("Invalid code. Please check and try again.");
+    }
+  };
+
+  return (
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/40">
```
ivantha: **[LOW]** Backdrop click does not close the modal; no Escape key support

The outer overlay `<div className="fixed inset-0 ...">` has no `onClick` handler, so clicking outside the modal card does nothing. Almost universally, users expect to close a modal by clicking the backdrop or pressing `Escape`. The `onClose` prop is available but never wired to these two standard dismissal paths.

**Suggestion:**
```
Wire up both:
```tsx
// Backdrop click
<div
  className="fixed inset-0 z-50 flex items-center justify-center bg-black/40"
  onClick={onClose}                // close on backdrop click
  onKeyDown={(e) => e.key === 'Escape' && onClose()}  // close on Escape
  role="presentation"
>
  <div
    className="bg-white rounded-xl ..."
    onClick={(e) => e.stopPropagation()}  // prevent card clicks from bubbling
  >
```
Or add a `useEffect(() => { const handler = (e) => e.key === 'Escape' && onClose(); document.addEventListener('keydown', handler); return () => document.removeEventListener('keydown', handler); }, [onClose]);` inside the modal component.
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx (hunk 2 above)
Comment (ivantha) [inline imagelab-frontend/src/components/SharePipelineModal.tsx]: **[LOW]** No unit tests for codec utility functions

The `compressToCode` and `decompressFromCode` functions are pure utility functions (modulo the Blockly workspace argument for compress) that implement a round-trip encode/decode contract. They are the most testable code in this PR but have no tests. The PR checklist confirms "New tests added" is unchecked. Edge cases like empty state, Unicode block names, and truncated/corrupted input are not guarded.

**Suggestion:**
```
Add unit tests for the codec:
```typescript
// SharePipelineModal.test.ts
describe('pipeline codec', () => {
  it('round-trips a pipeline state', () => {
    const state = { blocks: { languageVersion: 0, blocks: [] } };
    const code = btoa(/* encode */);
    expect(decompressFromCode(code)).toEqual(state);
  });

  it('throws on invalid base64 input', () => {
    expect(() => decompressFromCode('!!invalid!!')).toThrow();
  });

  it('throws on valid base64 but non-JSON content', () => {
    expect(() => decompressFromCode(btoa('not-json'))).toThrow();
  });
});
```
```

File: imagelab-frontend/src/components/Toolbar.tsx [hunk 5]
```diff
@@ -94,91 +98,106 @@ export default function Toolbar({ workspace }: ToolbarProps) {
...
+        </button>
+
+        <div className={separator} />
+
+        <button onClick={handleUndo} className={iconBtn} title={`Undo (${mod}Z)`}>
+          <Undo2 size={18} />
+        </button>
+        <button onClick={handleRedo} className={iconBtn} title={`Redo (${mod}Y or ${mod}⇧Z)`}>
+          <Redo2 size={18} />
+        </button>
```
ivantha: **[INFO]** Share button is not visually separated from the Run button

Looking at the toolbar layout, the existing pattern places a `<div className={separator} />` between each group of semantically related buttons: `[New | Download] | [Undo | Redo] | [Run]`. The new Share button is inserted after the Undo/Redo separator but before Run, with no separator between Share and Run. This groups a navigation/sharing action visually with the primary action (Run), which is inconsistent with the toolbar's existing grouping convention.

**Suggestion:**
```
Add a separator between Share and Run to maintain visual grouping:
```tsx
<button onClick={() => setShowShareModal(true)} ... >
  <Share2 size={18} />
</button>

<div className={separator} />  {/* ← add this */}

<button onClick={handleRun} ... >
```
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx (hunk 2 above)
Comment (ivantha) [inline imagelab-frontend/src/components/SharePipelineModal.tsx]: **[INFO]** Utility functions defined at module level could be extracted to a shared utilities file

`compressToCode` and `decompressFromCode` are module-level functions in the component file. This makes them harder to test in isolation, prevents reuse elsewhere (e.g., if a URL-share or export feature is added later), and adds cognitive weight to reading the component. The pattern in this codebase appears to use `hooks/` and `api/` directories for shared logic (e.g., `usePipeline`, `executePipeline`).

**Suggestion:**
```
Move to a dedicated utility module:
```
imagelab-frontend/src/utils/pipelineCodec.ts
```
```typescript
export function encodePipeline(workspace: Blockly.WorkspaceSvg): string { ... }
export function decodePipeline(code: string): unknown { ... }
```
Then import in the modal. This also makes the round-trip contract directly testable without needing to mount a React component.
```

File: imagelab-frontend/src/components/SharePipelineModal.tsx [hunk 6]
```diff
@@ -0,0 +1,196 @@
...
+      Blockly.serialization.workspaces.load(state, workspace);
+
+      const readImageBlocks = workspace.getBlocksByType("basic_readimage", false);
+      readImageBlocks.forEach((block) => {
+        const field = block.getField("filename_label");
+        if (field) field.setValue("No image");
+      });
+
+      setLoadSuccess(true);
+      setTimeout(() => onClose(), 800);
```
ivantha: **[INFO]** Auto-close after 800ms is an unusual and potentially disorienting UX pattern

After a successful load, the modal auto-closes after an 800ms delay via `setTimeout(() => onClose(), 800)`. This is an unconventional pattern — users may want to review the success message, or may be confused by a modal that disappears on its own. It also means if the user has started interacting with the newly-loaded workspace (since it's visible behind the modal), a delayed close could interrupt their action.

**Suggestion:**
```
Consider two alternatives:
1. **Close immediately** on success (simpler, and `loadSuccess` state is already set for the instant before close).
2. **Let the user dismiss** by changing the success UI to include a "Close" or "Done" button, or rely on the existing X button.

The 800ms delay is the worst of both approaches — too short for the user to read and decide, too long to feel instant.
```

ivantha (commented):
## ◎ Argus Review

This PR introduces a well-scoped Share Pipeline feature with a clean modal UI, thoughtful UX touches (copy feedback, disabled states, aria-labels), and a sensible round-trip encode/decode approach. However, there are two high-severity issues that need to be addressed before merge: **the workspace is silently wiped if loading fails** (data loss), and **there is no confirmation prompt before overwriting an existing workspace**. Additionally, the Clipboard API error path is unhandled, the modal is missing standard accessibility attributes (role, aria-modal, focus trapping), and the utility functions are misleadingly named since no compression actually occurs. The hardcoded block type and field name strings also create a silent failure point if block definitions are ever refactored.

PR #119, opened 2026-03-01
fix(conversions): handle grayscale input safely in GrayImage

### LINKED ISSUE #118: [Bug] GrayImage operator crashes on single-channel input
Title: GrayImage operator crashes on single-channel input [Bug]

## Problem

**`GrayImage` crashes when the input image is already grayscale (single-channel).**  
This breaks valid pipelines where a previous block already outputs a 1-channel image.

---

## Location
**Backend file:** `imagelab-backend/app/operators/conversions/gray_image.py`  
**Line:** ~9 (may shift as file changes)

---

## Root Cause

`GrayImage.compute()` always runs:

```python
cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)

### PR DESCRIPTION
**Pull request by Jayant-kernel**
## Description

Fixes the `GrayImage` conversion crash when input is already grayscale (single-channel). The operator now handles 1/3/4-channel inputs safely and raises a clear error for unsupported channel layouts. Added regression tests for grayscale, BGR, BGRA, and invalid 2-channel input.

Fixes #118

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added operator-level regression tests:
- `test_gray_image_returns_2d_grayscale_input_unchanged`
- `test_gray_image_converts_bgr_to_grayscale`
- `test_gray_image_converts_bgra_to_grayscale`
- `test_gray_image_handles_single_channel_3d_input`
- `test_gray_image_rejects_unsupported_channel_count`

Manual runtime validation done in backend venv with OpenCV installed:
- Grayscale input no longer crashes and remains unchanged
- BGR/BGRA paths match OpenCV expected output
- Unsupported channel count raises `ValueError`

- [ ] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/conversions/gray_image.py
+++ b/imagelab-backend/app/operators/conversions/gray_image.py
@@ -6,4 +6,16 @@ from app.operators.base import BaseOperator
 
 class GrayImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        return cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            return image
+
+        if image.ndim == 3:
+            channels = image.shape[2]
+            if channels == 1:
+                return image[:, :, 0]
+            if channels == 3:
+                return cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+            if channels == 4:
+                return cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
+
+        raise ValueError("GrayImage expects a 2D grayscale image or a 3D image with 1, 3, or 4 channels")
--- a/imagelab-backend/app/operators/conversions/gray_image.py
+++ b/imagelab-backend/app/operators/conversions/gray_image.py
@@ -6,16 +6,32 @@ from app.operators.base import BaseOperator
 
 class GrayImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        if image.ndim == 2:
-            return image
+        """
+        Convert *image* to a 2-D (H × W) grayscale array.
 
-        if image.ndim == 3:
+        Supported input formats
+        -----------------------
+        * 2-D ``(H, W)``         — already grayscale; returned as a copy.
+        * 3-D ``(H, W, 1)``      — single-channel packed; channel axis is squeezed, returned as a copy.
+        * 3-D ``(H, W, 3)``      — **BGR** (OpenCV convention) → grayscale.
+        * 3-D ``(H, W, 4)``      — **BGRA** (OpenCV convention) → grayscale; alpha ignored.
+
+        Raises
+        ------
+        ValueError
+            For any other dimensionality or channel count.
+        """
+        if image.ndim == 2:
+            return image.copy()
+        elif image.ndim == 3:
             channels = image.shape[2]
             if channels == 1:
-                return image[:, :, 0]
-            if channels == 3:
+                return image[:, :, 0].copy()
+            elif channels == 3:
                 return cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
-            if channels == 4:
+            elif channels == 4:
                 return cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
 
-        raise ValueError("GrayImage expects a 2D grayscale image or a 3D image with 1, 3, or 4 channels")
+        raise ValueError(
+            f"GrayImage expects a 2D or 3D image with 1/3/4 channels, got ndim={image.ndim}, shape={image.shape}"
+        )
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: imagelab-backend/app/operators/conversions/gray_image.py [hunk 1]
```diff
@@ -6,4 +6,16 @@
...
 class GrayImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        return cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            return image
+
+        if image.ndim == 3:
+            channels = image.shape[2]
+            if channels == 1:
+                return image[:, :, 0]
```
ivantha: **[MEDIUM]** Single-channel 3D slice returns a NumPy view, not a copy

`image[:, :, 0]` is a **view** into the original array's memory. Like the 2D case above, any in-place modification of the returned array propagates back to the original buffer.

This is subtly different from the BGR/BGRA branches, where `cv2.cvtColor` always allocates a fresh output array, making those branches safe. The inconsistency means the same operator behaves differently depending on the input format, which is unexpected and likely to surface as intermittent bugs in production.

**Suggestion:**
```
Append `.copy()` to make the slice independent:
```python
if channels == 1:
    return image[:, :, 0].copy()
```
```

File: imagelab-backend/tests/operators/conversions/test_gray_image.py [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+    assert np.array_equal(out, image[:, :, 0])
+
+
+def test_gray_image_rejects_unsupported_channel_count() -> None:
+    image = np.zeros((4, 4, 2), dtype=np.uint8)
+
+    with pytest.raises(
+        ValueError,
+        match="GrayImage expects a 2D grayscale image or a 3D image with 1, 3, or 4 channels",
+    ):
```
ivantha: **[INFO]** Exact string match in `pytest.raises` is fragile to message wording changes

The `match=` argument in `test_gray_image_rejects_unsupported_channel_count` uses the full, verbatim error message as a regex pattern:

```python
match="GrayImage expects a 2D grayscale image or a 3D image with 1, 3, or 4 channels"
```

If the error message is ever improved (e.g., to include the actual shape, as suggested above), this test will break even though the behaviour is correct. Pinning to the full string couples the test unnecessarily tightly to the message wording.

**Suggestion:**
```
Match only on a stable prefix or key phrase:
```python
with pytest.raises(ValueError, match=r"GrayImage expects"):
    GrayImage({}).compute(image)
```
```

File: imagelab-backend/app/operators/conversions/gray_image.py [hunk 3]
```diff
@@ -6,4 +6,16 @@
 
 class GrayImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        return cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            return image
```
ivantha: **[MEDIUM]** 2D input returned as-is, creating a mutable alias to the caller's array

When `image.ndim == 2`, the function returns the **same object** that was passed in — no copy is made. Any downstream code that writes to the returned array (e.g., normalising, blending, or running another operator in-place) will silently corrupt the upstream data held by the caller.

In image-processing pipelines this is a classic source of non-deterministic bugs that are extremely hard to reproduce: the issue only appears when a downstream stage modifies its input in-place, and only when the upstream stage was an already-gray image.

```python
source = load_gray_image()          # (H, W) ndarray
gray   = GrayImage({}).compute(source)
gray[0, 0] = 0                       # ← also zeroes source[0, 0]!
```

**Suggestion:**
```
Return a copy so operators are always independent:
```python
if image.ndim == 2:
    return image.copy()
```
```

ivantha (commented):
## ◎ Argus Review

This PR correctly fixes the crash that occurred when a grayscale (single-channel) image was passed to `GrayImage.compute()`, which previously called `cv2.cvtColor` unconditionally with `COLOR_BGR2GRAY`. The dispatch logic is sound, the error message is clear, and the five regression tests cover the primary cases well. However, the 2D and single-channel-3D branches return views/references into the original input array rather than copies, creating a mutation hazard that is typical of image-processing pipeline bugs and is not caught by the current test suite. A handful of lower-priority style, documentation, and coverage gaps round out the review.

Jayant-kernel:
Hi @ivantha, 
I've addressed all the review comments. All tests pass locally with uv run pytest -v and ruff is clean. Please take a look when you get a chance.

ivantha:

Jayant-kernel:
rebase done

PR #128, opened 2026-03-01
feat: add augmentation category with Gaussian and Salt & Pepper noise operators

### LINKED ISSUE #127: Feature - Add Augmentation category with Gaussian and Salt & Pepper Noise operators

Currently there is no way to simulate image noise in ImageLab. This makes it difficult to demonstrate de-noising techniques, since users cannot replicate common real world noise conditions such as sensor noise or corrupted pixel transmission directly within the pipeline.

A new Augmentation category in the sidebar with two operators - Gaussian Noise and Salt & Pepper Noise. Gaussian Noise adds random values sampled from a normal distribution to each pixel, simulating sensor noise. Salt & Pepper Noise randomly sets a proportion of pixels to either white or black, simulating corrupted pixels. Both operators are configurable via parameters exposed in the Blockly blocks.

The noise operators could have been placed under the existing Filtering category, however a dedicated Augmentation category is cleaner and more extensible for future contributions

**Additional context**

This is useful in an educational context: for example, a user can add Gaussian Noise to an image and then apply Gaussian Blur or Bilateral Filter to observe the de-noising effect, teaching the relationship between noise types and appropriate de-noising strategies.

I will be working on this issue.

xxivani:
## Description
Adds a new Augmentation category to the sidebar with two noise operators: Gaussian Noise and Salt & Pepper Noise. Gaussian Noise adds random values sampled from a normal distribution to each pixel, simulating sensor noise. Salt & Pepper Noise randomly sets a proportion of pixels to white or black, simulating corrupted pixel transmission. Both operators are fully configurable via Blockly block parameters and follow the existing full-stack pattern of backend operator, registry entry, block definition, category, theme style, and sidebar icon.

Fixes #127 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- Verified Gaussian Noise applies visible noise to the image with default and custom mean/sigma values.
- Verified Salt & Pepper Noise correctly corrupts a configurable proportion of pixels.
- Verified both blocks appear correctly in the new Augmentation category in the sidebar.
- Verified the pipeline runs end to end with both operators.
- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 
Salt and Pepper Noise
salt-pepper

Gaussian Noise
gaussian noise

## Checklist
- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/augmentation/gaussian_noise.py
+++ b/imagelab-backend/app/operators/augmentation/gaussian_noise.py
@@ -6,13 +6,29 @@ from app.operators.base import BaseOperator
 
 class GaussianNoise(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        mean = float(self.params.get("mean", 0))
-        sigma = float(self.params.get("sigma", 25))
+        try:
+            mean = float(self.params.get("mean", 0))
+        except (TypeError, ValueError):
+            mean = 0.0
+
+        try:
+            sigma = float(self.params.get("sigma", 25))
+        except (TypeError, ValueError):
+            sigma = 25.0
         sigma = max(1.0, sigma)
 
+        seed = self.params.get("seed", None)
+        rng = np.random.default_rng(int(seed) if seed is not None else None)
+
+        alpha = None
         if len(image.shape) == 3 and image.shape[2] == 4:
+            alpha = image[:, :, 3].copy()
             image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
 
-        noise = np.random.normal(mean, sigma, image.shape).astype(np.float32)
-        noisy = image.astype(np.float32) + noise
-        return np.clip(noisy, 0, 255).astype(np.uint8)
+        noise = rng.normal(mean, sigma, image.shape).astype(np.float32)
+        noisy = np.clip(image.astype(np.float32) + noise, 0, 255).astype(np.uint8)
+
+        if alpha is not None:
+            noisy = np.dstack([noisy, alpha])
+
+        return noisy
--- a/imagelab-backend/app/operators/augmentation/salt_pepper_noise.py
+++ b/imagelab-backend/app/operators/augmentation/salt_pepper_noise.py
@@ -6,28 +6,36 @@ from app.operators.base import BaseOperator
 
 class SaltPepperNoise(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
-        density = float(self.params.get("density", 0.05))
+        try:
+            density = float(self.params.get("density", 0.05))
+        except (TypeError, ValueError):
+            density = 0.05
         density = max(0.0, min(1.0, density))
 
+        seed = self.params.get("seed", None)
+        rng = np.random.default_rng(int(seed) if seed is not None else None)
+
+        alpha = None
         if len(image.shape) == 3 and image.shape[2] == 4:
+            alpha = image[:, :, 3].copy()
             image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
 
         result = image.copy()
         total_pixels = image.shape[0] * image.shape[1]
         num_affected = int(total_pixels * density)
+        num_affected -= num_affected % 2  # ensure even split
 
-        # Add salt (white pixels)
-        salt_coords = (
-            np.random.randint(0, image.shape[0], num_affected // 2),
-            np.random.randint(0, image.shape[1], num_affected // 2),
-        )
-        result[salt_coords] = 255
+        flat_indices = rng.choice(total_pixels, num_affected, replace=False)
+        salt_flat = flat_indices[: num_affected // 2]
+        pepper_flat = flat_indices[num_affected // 2 :]
 
-        # Add pepper (black pixels)
-        pepper_coords = (
-            np.random.randint(0, image.shape[0], num_affected // 2),
-            np.random.randint(0, image.shape[1], num_affected // 2),
-        )
-        result[pepper_coords] = 0
+        salt_rows, salt_cols = np.unravel_index(salt_flat, (image.shape[0], image.shape[1]))
+        pepper_rows, pepper_cols = np.unravel_index(pepper_flat, (image.shape[0], image.shape[1]))
+
+        result[salt_rows, salt_cols] = 255
+        result[pepper_rows, pepper_cols] = 0
+
+        if alpha is not None:
+            result = np.dstack([result, alpha])
 
         return result
--- a/imagelab-frontend/src/blocks/definitions/augmentation.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/augmentation.blocks.ts
@@ -4,9 +4,9 @@ export const augmentationBlocks = [
     message0: "Add Gaussian Noise %1 Mean %2 %3 Sigma %4",
     args0: [
       { type: "input_dummy" },
-      { type: "field_number", name: "mean", value: 0 },
+      { type: "field_number", name: "mean", value: 0, min: -127, max: 127 },
       { type: "input_dummy" },
-      { type: "field_number", name: "sigma", value: 25, min: 1 },
+      { type: "field_number", name: "sigma", value: 25, min: 1, max: 255 },
     ],
     previousStatement: null,
     nextStatement: null,


imagelab-backend/app/operators/augmentation/gaussian_noise.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class GaussianNoise(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        mean = float(self.params.get("mean", 0))
+        sigma = float(self.params.get("sigma", 25))
+        sigma = max(1.0, sigma)
+
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
```
ivantha: **[MEDIUM]** BGRA Alpha Channel Silently Dropped — Output Shape Changes

When the input image has 4 channels (BGRA), the code converts it to BGR with `cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)` and then returns a 3-channel result. The alpha channel is permanently discarded. If any downstream operator in the pipeline expects the same number of channels as the input (e.g., a write or blend operator), it will receive a shape mismatch and likely raise an error or produce incorrect output — silently, since no warning is emitted.

The same issue exists identically in `salt_pepper_noise.py` at lines 12–13.

**Suggestion:**
```
Preserve the alpha channel and re-attach it after processing:

```python
alpha = None
if len(image.shape) == 3 and image.shape[2] == 4:
    alpha = image[:, :, 3].copy()
    image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)

noise = np.random.normal(mean, sigma, image.shape).astype(np.float32)
noisy = np.clip(image.astype(np.float32) + noise, 0, 255).astype(np.uint8)

if alpha is not None:
    noisy = np.dstack([noisy, alpha])

return noisy
```

Apply the same fix to `salt_pepper_noise.py`.
```

File: imagelab-backend/app/operators/augmentation/salt_pepper_noise.py [hunk 2]
```diff
@@ -0,0 +1,33 @@
...
+from app.operators.base import BaseOperator
+
+
+class SaltPepperNoise(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        density = float(self.params.get("density", 0.05))
+        density = max(0.0, min(1.0, density))
+
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
```
ivantha: **[MEDIUM]** BGRA Alpha Channel Silently Dropped in Salt & Pepper Operator

Same issue as in `gaussian_noise.py`: when the input is a 4-channel BGRA image, the alpha channel is stripped during the `cv2.cvtColor` conversion and the returned array is BGR (3-channel). Any subsequent pipeline node that reads or composites the alpha channel will behave incorrectly. Because this is a silent shape change with no log or error, it can be difficult to diagnose.

**Suggestion:**
```
Apply the same alpha-preservation pattern described for `gaussian_noise.py`:

```python
alpha = None
if len(image.shape) == 3 and image.shape[2] == 4:
    alpha = image[:, :, 3].copy()
    image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)

result = image.copy()
# ... noise application ...

if alpha is not None:
    result = np.dstack([result, alpha])

return result
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR adds a well-structured Augmentation category with Gaussian Noise and Salt & Pepper Noise operators, following the project's full-stack pattern consistently across backend, block definitions, theme, and sidebar. However, there is a meaningful correctness bug in the Salt & Pepper implementation where independently sampled salt and pepper coordinates can overlap (causing pepper to silently overwrite salt pixels), and both operators silently drop the alpha channel of BGRA images without restoring it. No automated tests were added, and a few minor parameter-bounding gaps exist in the frontend block definitions. Overall the architecture is sound and the code is readable.

File: imagelab-backend/app/operators/augmentation/gaussian_noise.py [hunk 3]
```diff
@@ -0,0 +1,18 @@
+import cv2
+import numpy as np
+
+from app.operators.base import BaseOperator
+
+
+class GaussianNoise(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        mean = float(self.params.get("mean", 0))
+        sigma = float(self.params.get("sigma", 25))
```
ivantha: **[MEDIUM]** Unguarded `float()` Conversion Will Crash on Malformed Parameters

`float(self.params.get("mean", 0))` and `float(self.params.get("sigma", 25))` will raise an uncaught `ValueError` or `TypeError` if the pipeline passes a non-numeric string (e.g., `"abc"`, `None`, or a nested object). Depending on how the backend handles operator exceptions, this could return a 500 error without a meaningful message, or crash the pipeline worker. The same issue exists in `salt_pepper_noise.py` for the `density` parameter.

**Suggestion:**
```
Wrap parameter extraction in try/except and fall back to safe defaults:

```python
try:
    mean = float(self.params.get("mean", 0))
except (TypeError, ValueError):
    mean = 0.0

try:
    sigma = float(self.params.get("sigma", 25))
except (TypeError, ValueError):
    sigma = 25.0
sigma = max(1.0, sigma)
```

Alternatively, if there is a shared parameter-parsing utility in `BaseOperator` (e.g., `self.get_float_param("sigma", default=25, min=1.0)`), it should be used here for consistency.
```

File: imagelab-backend/app/operators/augmentation/gaussian_noise.py [hunk 4]
```diff
@@ -0,0 +1,18 @@
...
+        sigma = max(1.0, sigma)
+
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
+
+        noise = np.random.normal(mean, sigma, image.shape).astype(np.float32)
+        noisy = image.astype(np.float32) + noise
+        return np.clip(noisy, 0, 255).astype(np.uint8)
```
ivantha: **[MEDIUM]** No Automated Tests for Either New Operator

The PR checklist explicitly acknowledges that existing tests do not pass locally and no new tests were added. Both `GaussianNoise` and `SaltPepperNoise` are new public-facing operators with non-trivial logic (statistical behavior, BGRA conversion, clamping). Without tests:
- The coordinate-collision bug described above would have been caught.
- Regression protection for future refactors is absent.
- There is no way to verify that `sigma` clamping or `density` clamping works as intended.

**Suggestion:**
```
Add unit tests covering at minimum:

```python
# gaussian_noise_test.py
def test_output_shape_unchanged():
    img = np.zeros((100, 100, 3), dtype=np.uint8)
    op = GaussianNoise(params={"mean": 0, "sigma": 25})
    result = op.compute(img)
    assert result.shape == img.shape

def test_output_dtype_is_uint8():
    img = np.zeros((100, 100, 3), dtype=np.uint8)
    result = GaussianNoise(params={}).compute(img)

def test_output_values_clipped():
    img = np.full((50, 50, 3), 255, dtype=np.uint8)
    result = GaussianNoise(params={"mean": 100, "sigma": 50}).compute(img)
    assert result.max() <= 255 and result.min() >= 0

# salt_pepper_noise_test.py  
def test_density_zero_produces_unchanged_image():
    img = np.full((100, 100, 3), 128, dtype=np.uint8)
    result = SaltPepperNoise(params={"density": 0.0}).compute(img)
    np.testing.assert_array_equal(result, img)

def test_corrupted_pixel_count_is_approximate():
    img = np.full((200, 200, 3), 128, dtype=np.uint8)
    result = SaltPepperNoise(params={"density": 0.10}).compute(img)
    white = np.sum(np.all(result == 255, axis=2))
    black = np.sum(np.all(result == 0, axis=2))
    assert abs(white - black) < 500  # roughly equal salt and pepper
```
```

File: imagelab-frontend/src/blocks/definitions/augmentation.blocks.ts [hunk 5]
```diff
@@ -0,0 +1,37 @@
+export const augmentationBlocks = [
+  {
+    type: "augmentation_gaussiannoise",
+    message0: "Add Gaussian Noise %1 Mean %2 %3 Sigma %4",
+    args0: [
+      { type: "input_dummy" },
+      { type: "field_number", name: "mean", value: 0 },
```
ivantha: **[LOW]** Gaussian Noise `mean` Field Has No UI Bounds — Can Produce Degenerate Output

The `mean` `field_number` has no `min` or `max` constraints:
```ts
{ type: "field_number", name: "mean", value: 0 }
```
A user can type `mean = 5000`, and while the backend clips the result to `[0, 255]`, the image will be entirely white regardless of sigma. This can confuse learners who expect partial noise and instead see a completely white image. The backend's silent clipping hides the misconfiguration.

**Suggestion:**
```
Add reasonable bounds matching the pixel range. A mean of ±127 covers the full dynamic range without immediately washing out the image:

```ts
{ type: "field_number", name: "mean", value: 0, min: -127, max: 127 }
```

If intentional large offsets are a teaching goal, at minimum add a comment explaining the expected range and its effect.
```

File: imagelab-frontend/src/blocks/definitions/augmentation.blocks.ts [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+      { type: "input_dummy" },
+      { type: "field_number", name: "sigma", value: 25, min: 1 },
```
ivantha: **[LOW]** Gaussian Noise `sigma` Has No Upper Bound — Extreme Values Obscure the Image

`sigma` is constrained with `min: 1` but no `max`. A value like `sigma = 10000` generates noise that completely dominates the image signal; the output is visually indistinguishable from random noise regardless of the source image. While the backend correctly clips to `[0, 255]`, this extreme case is unlikely to be instructive and may look like a bug to users.

**Suggestion:**
```
Add a reasonable maximum. For a uint8 image, `sigma = 128` already produces very heavy noise; `sigma = 255` is the practical ceiling:

```ts
{ type: "field_number", name: "sigma", value: 25, min: 1, max: 255 }
```
```

File: imagelab-backend/app/operators/augmentation/gaussian_noise.py [hunk 7]
```diff
@@ -0,0 +1,34 @@
...
+            alpha = image[:, :, 3].copy()
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
+
+        noise = rng.normal(mean, sigma, image.shape).astype(np.float32)
+        noisy = np.clip(image.astype(np.float32) + noise, 0, 255).astype(np.uint8)
+
+        if alpha is not None:
+            noisy = np.dstack([noisy, alpha])
+
+        return noisy
```
ivantha: **[HIGH]** No automated tests added for new operators

Neither `GaussianNoise` nor `SaltPepperNoise` includes unit tests. The PR description's own checklist explicitly leaves "Existing tests pass", "New tests added", and "Tests pass locally" **unchecked**. This is especially important for numerical operators because edge cases (zero-size images, grayscale, BGRA, density=0, density=1, extreme sigma values) can silently produce wrong results or raise exceptions at runtime.

**Suggestion:**
```
Add pytest unit tests covering at minimum:
- BGR image (3-channel)
- Grayscale image (2D)
- BGRA image (4-channel, verify alpha is preserved)
- Edge-case parameters: `density=0`, `density=1`, `sigma=1`, `mean=-127`
- Reproducibility via `seed` parameter

```python
import numpy as np
from app.operators.augmentation.gaussian_noise import GaussianNoise

def test_gaussian_noise_bgr_shape_preserved():
    img = np.zeros((100, 100, 3), dtype=np.uint8)
    op = GaussianNoise(params={"mean": 0, "sigma": 25, "seed": 42})
    out = op.compute(img)
    assert out.shape == (100, 100, 3)
    assert out.dtype == np.uint8

def test_gaussian_noise_alpha_preserved():
    img = np.ones((50, 50, 4), dtype=np.uint8) * 128
    img[:, :, 3] = 200
    op = GaussianNoise(params={"mean": 0, "sigma": 25, "seed": 0})
    out = op.compute(img)
    assert out.shape == (50, 50, 4)
    np.testing.assert_array_equal(out[:, :, 3], 200)
```
```

File: imagelab-backend/app/operators/augmentation/gaussian_noise.py [hunk 8]
```diff
@@ -0,0 +1,34 @@
...
+            mean = 0.0
+
+        try:
+            sigma = float(self.params.get("sigma", 25))
+        except (TypeError, ValueError):
+            sigma = 25.0
+        sigma = max(1.0, sigma)
+
+        seed = self.params.get("seed", None)
+        rng = np.random.default_rng(int(seed) if seed is not None else None)
```
ivantha: **[MEDIUM]** seed type coercion raises ValueError for float-string inputs

`int(seed)` will raise `ValueError` if the seed value arrives as a float-string such as `"42.0"`, which is plausible if the Blockly front-end ever adds a seed field using `field_number`. The error is unhandled and propagates as a 500 response to the client. This contrasts with the careful try/except already applied to `mean`, `sigma`, and `density`.

**Suggestion:**
```
Use a try/except with a safe cast through float:

```python
# Before
rng = np.random.default_rng(int(seed) if seed is not None else None)

# After
try:
    rng_seed = int(float(seed)) if seed is not None else None
except (TypeError, ValueError):
    rng_seed = None
rng = np.random.default_rng(rng_seed)
```
```

File: imagelab-backend/app/operators/augmentation/salt_pepper_noise.py [hunk 9]
```diff
@@ -0,0 +1,41 @@
...
+class SaltPepperNoise(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        try:
+            density = float(self.params.get("density", 0.05))
+        except (TypeError, ValueError):
+            density = 0.05
+        density = max(0.0, min(1.0, density))
+
+        seed = self.params.get("seed", None)
+        rng = np.random.default_rng(int(seed) if seed is not None else None)
```
ivantha: **[MEDIUM]** seed type coercion raises ValueError for float-string inputs (salt & pepper)

Same issue as in `gaussian_noise.py`: `int(seed)` raises `ValueError` for float-string seeds. The rest of the parameter parsing in this file uses defensive try/except, but seed does not, creating an inconsistent robustness profile.

**Suggestion:**
```
```python
try:
    rng_seed = int(float(seed)) if seed is not None else None
except (TypeError, ValueError):
    rng_seed = None
rng = np.random.default_rng(rng_seed)
```
```

File: imagelab-backend/app/operators/augmentation/gaussian_noise.py [hunk 10]
```diff
@@ -0,0 +1,34 @@
...
+
+        alpha = None
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            alpha = image[:, :, 3].copy()
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
+
+        noise = rng.normal(mean, sigma, image.shape).astype(np.float32)
+        noisy = np.clip(image.astype(np.float32) + noise, 0, 255).astype(np.uint8)
```
ivantha: **[MEDIUM]** Gaussian noise sampled independently per channel diverges from documented behaviour

The noise array is generated with shape `image.shape` (e.g. `(h, w, 3)`), sampling **independent** noise for each colour channel. The PR description says it "adds values sampled from a Gaussian distribution to each pixel", which most users interpret as a single luminance perturbation applied uniformly across channels. With the current approach, individual channels receive different perturbations, so pixels can shift slightly in hue — a magenta or cyan tint appears in the noise — which is inconsistent with simulating sensor noise on a camera sensor (where luminance noise dominates).

**Suggestion:**
```
If per-luminance noise is intended, generate a single-channel map and broadcast:

```python
if len(image.shape) == 3:
    noise_2d = rng.normal(mean, sigma, image.shape[:2]).astype(np.float32)
    noise = np.stack([noise_2d] * image.shape[2], axis=-1)
else:
    noise = rng.normal(mean, sigma, image.shape).astype(np.float32)
noisy = np.clip(image.astype(np.float32) + noise, 0, 255).astype(np.uint8)
```
Alternatively, document the per-channel behaviour explicitly in the tooltip so users know what to expect.
```

File: imagelab-backend/app/operators/augmentation/salt_pepper_noise.py [hunk 11]
```diff
@@ -0,0 +1,41 @@
...
+        alpha = None
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            alpha = image[:, :, 3].copy()
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
+
+        result = image.copy()
+        total_pixels = image.shape[0] * image.shape[1]
+        num_affected = int(total_pixels * density)
+        num_affected -= num_affected % 2  # ensure even split
+
```
ivantha: **[LOW]** rng.choice without replacement is O(n) memory and slow at high density

`rng.choice(total_pixels, num_affected, replace=False)` uses a partial Fisher-Yates shuffle internally, allocating an array proportional to `total_pixels` regardless of `num_affected`. For a 4K image (~8.3 M pixels) at `density=0.5`, this creates a ~33 MB temporary index array and is measurably slower than a boolean mask approach.

**Suggestion:**
```
```python
# Boolean mask approach — faster and lower memory
mask = rng.random((image.shape[0], image.shape[1])) < density
split = rng.random((image.shape[0], image.shape[1])) < 0.5
salt_mask = mask & split
pepper_mask = mask & ~split
result[salt_mask] = 255
result[pepper_mask] = 0
```
This gives an approximately 50/50 salt/pepper split. If an exact 50/50 split is required, the current index-based approach can be kept but should be noted as a known performance trade-off.
```

File: imagelab-backend/app/operators/registry.py [hunk 12]
```diff
@@ -94,6 +96,9 @@
     "filtering_erosion": Erosion,
     "filtering_dilation": Dilation,
     "filtering_morphological": Morphological,
+    # Augmentation
+    "augmentation_gaussiannoise": GaussianNoise,
+    "augmentation_saltpeppernoise": SaltPepperNoise,
```
ivantha: **[LOW]** Augmentation entries inserted inside the Filtering section

The two new augmentation keys are placed between `filtering_morphological` and `filtering_gaborfilter`, interrupting the contiguous filtering block. This makes the registry harder to scan and could confuse future contributors who expect filtering entries to be contiguous.

**Suggestion:**
```
Move the augmentation block to its own clearly delineated section, either before the `# Filtering` block or after `# Thresholding`:

```python
    # Augmentation
    "augmentation_gaussiannoise": GaussianNoise,
    "augmentation_saltpeppernoise": SaltPepperNoise,
    # Filtering
    "filtering_blur": Blur,
    ...
```
```

PR #130, opened 2026-03-01
Feature - sepia filter

### LINKED ISSUE #129: Feature - Add Sepia Filter operator to Augmentation category

ImageLab does not yet support artistic or vintage color effects. Users have no operator to demonstrate color matrix transformations, which are a fundamental concept in image processing.

A Sepia Filter operator added to the Augmentation category. It applies a sepia color matrix to the image to produce a warm, brownish vintage tone. An intensity parameter (0–1) controls the strength of the effect: 0 leaves the image unchanged and 1 applies the full sepia transformation, with intermediate values blending between the original and the sepia result.

**Additional context**

This is useful for teaching color matrix transforms and blending operations: for eg, a user can experiment with different intensity values to understand how cv2.addWeighted blends two images together, making it a natural complement to the other operators in the Augmentation category.

xxivani:
## Description
Adds a Sepia Filter operator to the existing Augmentation category. The operator applies a sepia color matrix to the image to produce a warm, brownish vintage tone. An intensity parameter (0–1) controls the strength of the effect: 0 leaves the image unchanged and 1 applies the full sepia transformation.

Fixes #129 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- Verified sepia filter produces correct warm vintage tone at full intensity.
- Verified intensity parameter correctly blends between original and sepia at intermediate values.
- Verified intensity 0 returns the original image unchanged.
- Verified RGBA images are correctly converted to BGR before processing.
- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Screenshots 

Screenshot 2026-03-02 at 2 35 14 AM

## Checklist
- [ ] Tests pass locally

Note: This branch is based on feature/augmentation-operators. For a cleaner review of just the sepia-specific changes, comparing against that branch rather than main may be helpful.

### DIFF
--- a/imagelab-backend/app/operators/augmentation/sepia_filter.py
+++ b/imagelab-backend/app/operators/augmentation/sepia_filter.py
@@ -15,7 +15,10 @@ class SepiaFilter(BaseOperator):
     )
 
     def compute(self, image: np.ndarray) -> np.ndarray:
-        intensity = float(self.params.get("intensity", 1.0))
+        try:
+            intensity = float(self.params.get("intensity", 1.0))
+        except (TypeError, ValueError):
+            intensity = 1.0
         intensity = max(0.0, min(1.0, intensity))
 
         if intensity == 0.0:
@@ -23,8 +26,8 @@ class SepiaFilter(BaseOperator):
 
         alpha = None
         if len(image.shape) == 3 and image.shape[2] == 4:
-            alpha = image[:, :, 3]
-            image = image[:, :, :3]
+            alpha = image[:, :, 3].copy()
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
         elif len(image.shape) == 2:
             image = cv2.cvtColor(image, cv2.COLOR_GRAY2BGR)
 
@@ -32,7 +35,7 @@ class SepiaFilter(BaseOperator):
         sepia_f = cv2.transform(img_f, self._SEPIA_MATRIX)
         sepia_f = np.clip(sepia_f, 0, 255)
 
-        result = np.clip((1.0 - intensity) * img_f + intensity * sepia_f, 0, 255).astype(np.uint8)
+        result = ((1.0 - intensity) * img_f + intensity * sepia_f).astype(np.uint8)
 
         if alpha is not None:
             result = cv2.merge([result[:, :, 0], result[:, :, 1], result[:, :, 2], alpha])


imagelab-backend/app/operators/augmentation/sepia_filter.py
```diff
@@ -0,0 +1,30 @@
...
+                [0.168, 0.686, 0.349],
+                [0.189, 0.769, 0.393],
+            ],
+            dtype=np.float32,
+        )
+
+        sepia = cv2.transform(image.astype(np.float32), sepia_matrix)
+        sepia = np.clip(sepia, 0, 255).astype(np.uint8)
+
+        result = cv2.addWeighted(image, 1.0 - intensity, sepia, intensity, 0)
```
ivantha: **[LOW]** Full sepia transform computed even when `intensity == 0.0`

When `intensity` is `0.0`, the result of `addWeighted` is the unmodified original image (weight `1.0`) with the sepia output discarded (weight `0.0`). Despite this, the code still:
1. Calls `image.astype(np.float32)` — allocates a full float32 copy of the image.
2. Calls `cv2.transform` — performs the matrix multiplication across all pixels.
3. Calls `np.clip(...).astype(np.uint8)` — allocates another full copy.

All three operations are O(H×W) and produce output that is immediately discarded.

**Suggestion:**
```
Short-circuit at `intensity == 0.0` (or a small epsilon):

```python
if intensity == 0.0:
    return image.copy()  # return a copy for consistency
```

A similar guard for `intensity == 1.0` could also skip the `addWeighted` call, though that is more of a micro-optimisation.
```

File: imagelab-backend/app/operators/augmentation/sepia_filter.py [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+
+        # Sepia color matrix
+        sepia_matrix = np.array(
+            [
+                [0.131, 0.534, 0.272],
+                [0.168, 0.686, 0.349],
+                [0.189, 0.769, 0.393],
+            ],
+            dtype=np.float32,
+        )
```
ivantha: **[INFO]** Sepia matrix is correctly formulated for OpenCV BGR channel order — well done

This is a positive observation. The standard sepia transformation is typically documented for **RGB** input:
```
R_out = 0.393·R + 0.769·G + 0.189·B
G_out = 0.349·R + 0.686·G + 0.168·B
B_out = 0.272·R + 0.534·G + 0.131·B
```
The author has correctly **transposed** the channel ordering for OpenCV's BGR convention. Verifying row-by-row against `cv2.transform`'s `dst[i] = m[i,·] · src` semantics with `src = [B, G, R]`:
- Row 0 (→ B_out): `0.131·B + 0.534·G + 0.272·R` ✓  
- Row 1 (→ G_out): `0.168·B + 0.686·G + 0.349·R` ✓  
- Row 2 (→ R_out): `0.189·B + 0.769·G + 0.393·R` ✓  

This is easy to get wrong (many online examples use the RGB form directly, producing a blue-tinted sepia when applied to BGR images), so it is notable that it is correct here.

**Suggestion:**
```
Consider adding a comment to the matrix definition to make the channel ordering explicit for future maintainers:

```python
# Sepia matrix for BGR input (transposed from the standard RGB form).
# Row i = coefficients [B_weight, G_weight, R_weight] for output channel i.
sepia_matrix = np.array(
    [
        [0.131, 0.534, 0.272],  # B_out
        [0.168, 0.686, 0.349],  # G_out
        [0.189, 0.769, 0.393],  # R_out
    ],
    dtype=np.float32,
)
```
```

File: imagelab-backend/app/operators/augmentation/sepia_filter.py [hunk 3]
```diff
@@ -0,0 +1,30 @@
...
+
+        sepia = cv2.transform(image.astype(np.float32), sepia_matrix)
+        sepia = np.clip(sepia, 0, 255).astype(np.uint8)
+
+        result = cv2.addWeighted(image, 1.0 - intensity, sepia, intensity, 0)
+
+        return result
```
ivantha: **[INFO]** Consistent operator structure across all three new classes — good pattern

All three operators follow a uniform, readable pattern:
1. Extract and coerce parameters from `self.params` with safe defaults.
2. Apply bounds clamping where appropriate.
3. Normalise input channel count (BGRA → BGR guard).
4. Perform the core computation.
5. Return a `uint8` result.

This is consistent with the existing codebase style and makes the operators easy to understand, extend, and review. The frontend block definitions and registry keys also match cleanly, reducing the risk of name/type mismatches.

**Suggestion:**
```
No change required. Consider encoding this pattern as a `BaseOperator` contract or mixin in the future (e.g., a `normalise_input()` helper that centralises the BGRA/grayscale handling so it doesn't need to be duplicated in every operator).
```

PR #142, opened 2026-03-04
test: add unit tests for thresholding operators

### LINKED ISSUE #141: Test - Add unit tests for thresholding operators

The thresholding operators (ApplyThreshold, AdaptiveThreshold, OtsuThreshold) have no unit tests. This means regressions in their behaviour, such as incorrect default values, wrong output dtype, or broken grayscale/RGBA handling cannot be caught automatically.

Add a unit test suite covering all three thresholding operators, including shape and dtype assertions, binary output verification, grayscale and RGBA input handling, default parameter behaviour, and the existing maxValue=0 bug in ApplyThreshold.

xxivani:
## Description
Adds unit tests for the three thresholding operators: ApplyThreshold, AdaptiveThreshold, and OtsuThreshold. Tests cover output shape, dtype, binary output correctness, colour and RGBA input auto-conversion, even block size correction in AdaptiveThreshold, and a confirming test for the default maxValue=0 bug in ApplyThreshold which produces an all-black output.

To run the tests:
```
cd imagelab-backend
uv sync --extra dev
uv run pytest tests/test_thresholding_operators.py -v
```

Fixes #141 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

All 21 tests pass locally via `uv run pytest tests/test_thresholding_operators.py -v`.

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/tests/conftest.py
+++ b/imagelab-backend/tests/conftest.py
@@ -49,18 +49,6 @@ def png_b64() -> str:
     return base64.b64encode(buf.tobytes()).decode()
 
 
-@pytest.fixture
-def color_image():
-    rng = np.random.default_rng(42)
-    return rng.integers(0, 256, (100, 100, 3), dtype=np.uint8)
-
-
-@pytest.fixture
-def grayscale_image():
-    rng = np.random.default_rng(42)
-    return rng.integers(0, 256, (100, 100), dtype=np.uint8)
-
-
 @pytest.fixture(scope="session")
 def sample_image_b64() -> str:
     img = np.full((100, 100, 3), 255, dtype=np.uint8)
--- a/imagelab-backend/tests/test_thresholding_operators.py
+++ b/imagelab-backend/tests/test_thresholding_operators.py
@@ -1,10 +1,9 @@
 import numpy as np
+import pytest
 
 from app.operators.thresholding.adaptive_threshold import AdaptiveThreshold
 from app.operators.thresholding.apply_threshold import ApplyThreshold
 from app.operators.thresholding.otsu_threshold import OtsuThreshold
-
-# ApplyThreshold
 
 
 class TestApplyThreshold:
@@ -16,25 +15,36 @@ class TestApplyThreshold:
         result = ApplyThreshold({}).compute(color_image)
         assert result.dtype == np.uint8
 
-    def test_default_max_value_produces_black_output(self, color_image):
+    @pytest.mark.xfail(
+        strict=True, reason="maxValue defaults to 0 — known bug, fix in fix/apply-threshold-default-max-value"
+    )
+    def test_default_params_produces_non_empty_output(self, color_image):
         result = ApplyThreshold({}).compute(color_image)
-        assert result.max() == 0
+        assert result.max() > 0
 
     def test_max_value_255_produces_binary_output(self, grayscale_image):
         result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(grayscale_image)
         unique = np.unique(result)
         assert set(unique).issubset({0, 255})
 
-    def test_threshold_value_0_with_max_255_sets_all_pixels(self, grayscale_image):
+    def test_threshold_value_0_pixels_above_zero_become_max(self, grayscale_image):
         result = ApplyThreshold({"maxValue": 255, "thresholdValue": 0}).compute(grayscale_image)
-        assert result.max() == 255
+        assert np.all(result[grayscale_image > 0] == 255)
+        assert np.all(result[grayscale_image == 0] == 0)
 
     def test_grayscale_input(self, grayscale_image):
         result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(grayscale_image)
         assert result.shape == grayscale_image.shape
 
+    def test_color_input_output_channels(self, color_image):
+        result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(color_image)
+        assert result.shape == color_image.shape
 
-# AdaptiveThreshold
+    def test_threshold_at_boundary(self):
+        img = np.array([[0, 127, 128, 255]], dtype=np.uint8)
+        result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(img)
+        expected = np.array([[0, 0, 255, 255]], dtype=np.uint8)
+        np.testing.assert_array_equal(result, expected)
 
 
 class TestAdaptiveThreshold:
@@ -77,8 +87,10 @@ class TestAdaptiveThreshold:
         result = AdaptiveThreshold({}).compute(grayscale_image)
         assert result.shape == grayscale_image.shape
 
-
-# OtsuThreshold
+    def test_invalid_adaptive_method_falls_back_to_gaussian(self, color_image):
+        result_invalid = AdaptiveThreshold({"adaptiveMethod": "INVALID"}).compute(color_image)
+        result_gaussian = AdaptiveThreshold({"adaptiveMethod": "GAUSSIAN"}).compute(color_image)
+        np.testing.assert_array_equal(result_invalid, result_gaussian)
 
 
 class TestOtsuThreshold:
@@ -99,6 +111,12 @@ class TestOtsuThreshold:
         result = OtsuThreshold({}).compute(color_image)
         assert len(result.shape) == 2
 
+    def test_rgba_input(self, color_image):
+        rgba = np.dstack([color_image, np.full(color_image.shape[:2], 255, dtype=np.uint8)])
+        result = OtsuThreshold({}).compute(rgba)
+        assert len(result.shape) == 2
+        assert set(np.unique(result)).issubset({0, 255})
+
     def test_grayscale_input(self, grayscale_image):
         result = OtsuThreshold({}).compute(grayscale_image)
         assert result.shape == grayscale_image.shape
@@ -107,3 +125,14 @@ class TestOtsuThreshold:
         result = OtsuThreshold({"maxValue": 128}).compute(grayscale_image)
         unique = np.unique(result)
         assert set(unique).issubset({0, 128})
+
+    def test_otsu_threshold_adapts_to_image_content(self):
+        dark_img = np.zeros((100, 100), dtype=np.uint8)
+        dark_img[:80, :] = 20
+        dark_img[80:, :] = 230
+        bright_img = np.zeros((100, 100), dtype=np.uint8)
+        bright_img[:20, :] = 20
+        bright_img[20:, :] = 230
+        result_dark = OtsuThreshold({"maxValue": 255}).compute(dark_img)
+        result_bright = OtsuThreshold({"maxValue": 255}).compute(bright_img)
+        assert result_dark.mean() != result_bright.mean()


ivantha:
## ◎ Argus Review

This PR adds 21 unit tests for three thresholding operators (ApplyThreshold, AdaptiveThreshold, OtsuThreshold). The test structure is clean, fixtures use modern NumPy APIs with fixed seeds for reproducibility, and edge-case coverage (RGBA input, even block-size correction) is commendable. However, the most critical concern is `test_default_max_value_produces_black_output`, which deliberately codifies a known bug as an assertion — encoding broken behaviour as a passing test makes the bug harder to fix and gives false confidence. Several other findings relate to weak assertions, missing consistency between test classes, and missing `import pytest`.

### Findings (12)

#### 🟠 Bug-confirming test codifies known broken behaviour as a passing assertion
**Severity:** high | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:19`

The test `test_default_max_value_produces_black_output` asserts that `result.max() == 0`, which the PR description itself acknowledges is caused by a **bug** (`maxValue` defaulting to `0`). Merging a green test that encodes incorrect behaviour is actively harmful:

1. **Blocks the bug fix** – any future PR that corrects the default will immediately break this test, creating friction without adding safety.
2. **Misleads CI** – the test suite goes green while a real bug exists, giving false confidence.
3. **Ambiguous intent** – a future maintainer reading this test cannot easily tell whether the all-black output is a design decision or a defect.

The correct approach depends on the team's workflow:
- If the bug **will be fixed in a follow-up PR**, mark the test with `pytest.mark.xfail(strict=True, reason="maxValue defaults to 0 — issue #141")` and assert the *correct* outcome (e.g., `result.max() == 255`).
- If the fix is **in scope for this PR**, fix the default and write a test for the correct behaviour.
- If the test is meant purely as regression documentation, add a prominent comment explaining it tests a *known bug* and reference the issue.

**Suggestion:**
```
```python
# Option A – xfail documents the bug and will auto-promote to a real test once fixed
@pytest.mark.xfail(strict=True, reason="Default maxValue=0 is a bug (issue #141); expected result should not be all-black")
def test_default_max_value_bug(self, color_image):
    result = ApplyThreshold({}).compute(color_image)
    # Correct behaviour: at least some pixels should be non-zero with a sane default
    assert result.max() > 0

# Option B – fix the bug and write the positive test
def test_default_params_produces_non_empty_output(self, color_image):
    result = ApplyThreshold({}).compute(color_image)
    assert result.max() > 0
```
```

---

#### 🟡 Misleading test name and weak assertion in `test_threshold_value_0_with_max_255_sets_all_pixels`
**Severity:** medium | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:28`

The test name says "sets **all** pixels" to 255, but `assert result.max() == 255` only confirms that **at least one** pixel reached 255. With `THRESH_BINARY` and `thresholdValue=0`, pixels with value **exactly 0** remain 0 (the condition is `pixel_value > threshold`, not `>=`). `rng.integers(0, 256, …)` can produce zeros, so the claim that "all pixels" are set is incorrect both semantically and in the assertion.

The assertion is too weak to catch regressions where, say, only half the pixels are correctly thresholded.

**Suggestion:**
```
```python
def test_threshold_value_0_with_max_255_most_pixels_set(self, grayscale_image):
    result = ApplyThreshold({"maxValue": 255, "thresholdValue": 0}).compute(grayscale_image)
    # Pixels strictly > 0 must become 255; pixels == 0 remain 0
    assert np.all(result[grayscale_image > 0] == 255)
    assert np.all(result[grayscale_image == 0] == 0)
```
```

---

#### 🟡 ApplyThreshold shape contract is inconsistent with the other two operators — and untested for grayscale output
**Severity:** medium | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:11`

`TestAdaptiveThreshold` and `TestOtsuThreshold` both assert `result.shape == color_image.shape[:2]` (2-D output), while `TestApplyThreshold.test_default_params_output_shape` asserts `result.shape == color_image.shape` (preserving the 3-channel dimension). This suggests one of two things:

1. **ApplyThreshold intentionally thresholds each channel independently** — if so, this is a notable API difference that deserves a docstring or comment explaining the design, and there should be a test that verifies per-channel correctness.
2. **ApplyThreshold silently converts to grayscale but then broadcasts back** — if so, the shape-equality assertion is accidentally passing for a wrong reason.

Either way, there is no test that verifies the *content* of a multi-channel thresholded result, making it easy for subtle bugs to slip through.

**Suggestion:**
```
Add an explicit test that asserts and documents the expected channel behaviour:
```python
def test_color_input_output_channels(self, color_image):
    result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(color_image)
    # Document whether output is 2-D (grayscale) or 3-D (per-channel)
    assert result.ndim == 2, "ApplyThreshold should reduce to single-channel output"
    # OR
    assert result.shape == color_image.shape, "ApplyThreshold applies threshold per-channel"
```
Also add a short docstring to `ApplyThreshold.compute` explaining the channel contract.
```

---

#### 🟡 Missing RGBA input test for OtsuThreshold — inconsistent coverage vs AdaptiveThreshold
**Severity:** medium | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:84`

`TestAdaptiveThreshold` includes `test_rgba_input` (lines 58-61) to verify that 4-channel images are handled, but `TestOtsuThreshold` has no equivalent test. If both operators share the same grayscale-conversion code path, the RGBA case should be equally important for Otsu. If the code paths differ, the missing test is a gap that could hide a bug with RGBA inputs.

**Suggestion:**
```
```python
def test_rgba_input(self, color_image):
    rgba = np.dstack([color_image, np.full(color_image.shape[:2], 255, dtype=np.uint8)])
    result = OtsuThreshold({}).compute(rgba)
    assert len(result.shape) == 2
    assert set(np.unique(result)).issubset({0, 255})
```
```

---

#### 🟡 `test_even_block_size_corrected_to_odd` documents silent coercion without questioning the design
**Severity:** medium | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:63`

The test verifies that passing an even `blockSize=4` silently produces identical results to `blockSize=5`. While this is useful regression coverage, the test legitimises a silent data-mutation pattern that can be surprising to callers. An even `blockSize` is an invalid argument to `cv2.adaptiveThreshold`; silently rounding up masks misconfiguration errors at the call site.

From a review standpoint, this raises the question: **should the operator warn or raise instead of silently correcting?** A `logging.warning` or even a `ValueError` with a helpful message ("blockSize must be odd; received 4") would be more debuggable than silent coercion. The test as written gives the impression this is accepted, desirable behaviour.

**Suggestion:**
```
Consider whether the operator should warn:
```python
import warnings

if block_size % 2 == 0:
    warnings.warn(
        f"blockSize must be odd; {block_size} was automatically rounded up to {block_size + 1}.",
        UserWarning,
        stacklevel=2,
    )
    block_size += 1
```
Then update the test:
```python
def test_even_block_size_corrected_to_odd(self, color_image):
    with pytest.warns(UserWarning, match="blockSize must be odd"):
        result_even = AdaptiveThreshold({"blockSize": 4}).compute(color_image)
    result_odd = AdaptiveThreshold({"blockSize": 5}).compute(color_image)
    np.testing.assert_array_equal(result_even, result_odd)
```
```

---

#### 🔵 Missing `import pytest` limits future test extensibility
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:1`

The file has no `import pytest` statement. Currently this is not strictly needed (no `pytest.mark`, `pytest.raises`, or `pytest.warns` is used), but it means any future test improvements — such as `pytest.mark.xfail` for the bug-documenting test, `pytest.raises` for error-path tests, or `pytest.warns` for the block-size coercion warning — require adding the import. Given that such additions are likely (see other findings), the import should be present from the start.

**Suggestion:**
```
```python
import pytest
import numpy as np

from app.operators.thresholding.adaptive_threshold import AdaptiveThreshold
from app.operators.thresholding.apply_threshold import ApplyThreshold
from app.operators.thresholding.otsu_threshold import OtsuThreshold
```
```

---

#### 🔵 Fixtures default to `function` scope — `module` scope would reduce redundant allocations
**Severity:** low | **Category:** performance | **File:** `imagelab-backend/tests/conftest.py:37`

Both `color_image` and `grayscale_image` return fresh NumPy arrays on every test invocation (default `scope="function"`). Since no test mutates the fixture value, the arrays are safe to share within a module. For 100×100 images and 21 tests this is negligible, but establishing the habit of using `scope="module"` for read-only fixtures keeps the pattern correct as test suites grow.

**Suggestion:**
```
```python
@pytest.fixture(scope="module")
def color_image():
    rng = np.random.default_rng(42)
    return rng.integers(0, 256, (100, 100, 3), dtype=np.uint8)

@pytest.fixture(scope="module")
def grayscale_image():
    rng = np.random.default_rng(42)
    return rng.integers(0, 256, (100, 100), dtype=np.uint8)
```
If you switch to `module` scope, ensure that no test mutates the arrays in place (e.g., no `image[:] = …`). If tests do mutate, keep `function` scope or return `.copy()` in each fixture.
```

---

#### 🔵 No test verifies Otsu's automatic threshold selection property
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:84`

OtsuThreshold's defining characteristic is that it **automatically computes** an optimal threshold from the image histogram. The current tests only check output shape, dtype, and binary values — all of which would pass even if a fixed threshold (e.g., 127) were used instead of Otsu's algorithm. There is no test that demonstrates the threshold adapts to image content, which means the Otsu-specific logic could be silently replaced with a fixed-value threshold and all tests would still pass.

**Suggestion:**
```
```python
def test_otsu_threshold_adapts_to_image_content(self):
    """Otsu threshold must differ for images with very different bimodal distributions."""
    rng = np.random.default_rng(0)
    # Dark image: most pixels near 0
    dark = np.concatenate([rng.integers(0, 30, 5000), rng.integers(200, 256, 5000)])
    # Bright image: most pixels near 255 (inverted distribution)
    bright = np.concatenate([rng.integers(0, 56, 5000), rng.integers(200, 256, 5000)])
    dark_img = dark.reshape(100, 100).astype(np.uint8)
    bright_img = bright.reshape(100, 100).astype(np.uint8)
    # The fraction of pixels thresholded to 255 should differ
    result_dark = OtsuThreshold({"maxValue": 255}).compute(dark_img)
    result_bright = OtsuThreshold({"maxValue": 255}).compute(bright_img)
    assert result_dark.mean() != result_bright.mean()
```
```

---

#### 🔵 No error-path or invalid-input tests across all three operator classes
**Severity:** low | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:10`

All 21 tests exercise the happy path with valid inputs. There are no tests for:
- Invalid `thresholdType` or `adaptiveMethod` string values
- Float32 / float64 input images (OpenCV thresholding typically requires uint8)
- Images with zero dimension or a single pixel
- `blockSize=1` for AdaptiveThreshold (minimum valid odd value; may behave oddly)

Without negative tests, the operators' error-handling code (if any) goes completely untested.

**Suggestion:**
```
```python
# Example: test that an unsupported dtype raises a clear error
def test_float_input_raises_or_converts(self, grayscale_image):
    float_img = grayscale_image.astype(np.float32) / 255.0
    with pytest.raises((ValueError, cv2.error)):
        ApplyThreshold({"maxValue": 255, "thresholdValue": 0.5}).compute(float_img)

# Example: test invalid adaptiveMethod
def test_invalid_adaptive_method_raises(self, color_image):
    with pytest.raises((ValueError, KeyError)):
        AdaptiveThreshold({"adaptiveMethod": "INVALID"}).compute(color_image)
```
```

---

#### ℹ️ No module-level docstring or inline comments explaining test rationale
**Severity:** info | **Category:** documentation | **File:** `imagelab-backend/tests/test_thresholding_operators.py:1`

The file has no module-level docstring summarising what is being tested and why. Several tests—particularly `test_default_max_value_produces_black_output`, `test_even_block_size_corrected_to_odd`, and `test_threshold_value_0_with_max_255_sets_all_pixels`—test non-obvious behaviour. Without inline comments, future readers cannot distinguish intentional edge-case documentation from accidental assertions.

**Suggestion:**
```
Add a brief module docstring and comments for non-obvious tests:
```python
"""
Unit tests for the thresholding operator family:
  - ApplyThreshold  (simple binary threshold)
  - AdaptiveThreshold (local neighbourhood threshold)
  - OtsuThreshold   (automatic global threshold via Otsu's method)

All tests use deterministic random images (seed=42) to ensure reproducibility.
Note: test_default_max_value_produces_black_output documents a known bug (issue #141).
"""
```
```

---

#### ℹ️ Both fixtures share seed 42 — edge cases like all-zero rows are not exercised
**Severity:** info | **Category:** testing | **File:** `imagelab-backend/tests/conftest.py:37`

Using the same seed for both fixtures is fine for determinism, but the 100×100 random images are unlikely to hit edge cases (e.g., all pixels equal to the threshold value, a row of zeros, or a uniform image). Consider complementing the random-image fixtures with a few hand-crafted images in individual tests where boundary conditions are being exercised (e.g., `test_threshold_value_0_with_max_255_sets_all_pixels` would benefit from an image that explicitly includes zero-valued pixels).

**Suggestion:**
```
For boundary-condition tests, construct purpose-built images inline:
```python
def test_threshold_at_boundary(self):
    # Image with known pixel values straddling the threshold
    img = np.array([[0, 127, 128, 255]], dtype=np.uint8)
    result = ApplyThreshold({"maxValue": 255, "thresholdValue": 127}).compute(img)
    expected = np.array([[0, 0, 255, 255]], dtype=np.uint8)  # strict >
    np.testing.assert_array_equal(result, expected)
```
```

---

#### ℹ️ Missing tests for threshold type variants across all operators
**Severity:** info | **Category:** testing | **File:** `imagelab-backend/tests/test_thresholding_operators.py:23`

OpenCV's thresholding supports multiple modes: `THRESH_BINARY`, `THRESH_BINARY_INV`, `THRESH_TRUNC`, `THRESH_TOZERO`, `THRESH_TOZERO_INV`. None of the 21 tests exercise a non-default `thresholdType`, meaning the parameter-forwarding code for these variants is untested. A typo or wrong constant in the mapping would go undetected.

**Suggestion:**
```
```python
def test_inverted_binary_threshold(self, grayscale_image):
    result = ApplyThreshold({
        "maxValue": 255,
        "thresholdValue": 127,
        "thresholdType": "BINARY_INV"
    }).compute(grayscale_image)
    unique = np.unique(result)
    assert set(unique).issubset({0, 255})
    # Pixels below threshold should now be 255
    assert np.all(result[grayscale_image <= 127] == 255)
```
```

---

PR #143, opened 2026-03-04
fix: enable real-time snap and ghost for sidebar blocks

### LINKED ISSUE #139: Blocks do not snap in real-time while dragging

Yes. When dragging a Blockly block to connect it to another block, the block does not snap or stick instantly while being dragged. I have to release the block first for it to connect, which makes the dragging experience feel delayed and less intuitive.

The block should snap / stick to the target block in real time during the drag operation, so that the connection is visible immediately without needing to release the block.

None.

**Additional context**

https://github.com/user-attachments/assets/e29d737c-977c-442f-ab93-e50b71ce18ed

MohamedAli1937:
## Description

Implemented real-time block snapping and ghost preview for blocks dragged from the sidebar. Now blocks snap immediately when dropped, and all dragged blocks show consistent visual feedback similar to existing workspace blocks.

Fixes #139 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

https://github.com/user-attachments/assets/e0a28965-74e5-4667-852f-a083c02f4ca5

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-frontend/src/hooks/useBlocklyWorkspace.ts
+++ b/imagelab-frontend/src/hooks/useBlocklyWorkspace.ts
@@ -14,6 +14,14 @@ import {
 } from "./workspacePersistence";
 
 const SAVE_DEBOUNCE_MS = 500;
+
+const SNAP_RADIUS = 48;
+const CONNECTING_SNAP_RADIUS = 68;
+
+// Apply global Blockly configuration once at module load
+Blockly.config.snapRadius = SNAP_RADIUS;
+Blockly.config.connectingSnapRadius = CONNECTING_SNAP_RADIUS;
+
 const MUTATING_EVENTS = new Set<string>([
   Blockly.Events.BLOCK_CREATE,
   Blockly.Events.BLOCK_DELETE,
@@ -33,6 +41,9 @@ export function useBlocklyWorkspace() {
 
   const initWorkspace = useCallback(() => {
     if (!containerRef.current || workspaceRef.current) return;
+
+    Blockly.config.snapRadius = 48;
+    Blockly.config.connectingSnapRadius = 68;
 
     const ws = Blockly.inject(containerRef.current, {
       readOnly: false,
--- a/imagelab-frontend/src/hooks/useSidebarDrag.ts
+++ b/imagelab-frontend/src/hooks/useSidebarDrag.ts
@@ -4,11 +4,30 @@ import { SINGLETON_BLOCK_TYPES, isSingletonBlockPresent } from "../utils/blockLi
 
 const DRAG_THRESHOLD = 5;
 
+function createSyntheticPointerEvent(e: MouseEvent): PointerEvent {
+  return new PointerEvent("pointermove", {
+    clientX: e.clientX,
+    clientY: e.clientY,
+    bubbles: true,
+    pointerId: 1,
+  });
+}
+
 interface UseSidebarDragOptions {
   type: string;
   label: string;
   workspace: Blockly.WorkspaceSvg | null;
   previewDataUrl?: string;
+}
+
+export function isOverWorkspace(
+  rect: DOMRect,
+  clientX: number,
+  clientY: number,
+): boolean {
+  return (
+    clientX >= rect.left && clientX <= rect.right && clientY >= rect.top && clientY <= rect.bottom
+  );
 }
 
 export function useSidebarDrag({ type, label, workspace, previewDataUrl }: UseSidebarDragOptions) {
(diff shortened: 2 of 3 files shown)


imagelab-frontend/src/hooks/useSidebarDrag.ts
```diff
@@ -32,78 +98,151 @@ export function useSidebarDrag({ type, label, workspace, previewDataUrl }: UseSi
...
+        const dy = e.clientY - startPos.current.y;
+
+        if (!isDragging.current && Math.sqrt(dx * dx + dy * dy) > DRAG_THRESHOLD) {
+          isDragging.current = true;
+          wasDragged.current = true;
+
+          const ghost = document.createElement("div");
+          ghost.className = "sidebar-drag-ghost";
+          ghost.style.pointerEvents = "none";
+          if (previewDataUrl) {
```
ivantha: **[HIGH]** Stale workspace bounds cache not invalidated on scroll

The `wsBoundsRef` is lazily populated on the first `mousemove` over the workspace and only cleared by a `resize` event. However, if the page or any ancestor element scrolls, the cached `DOMRect` becomes stale. This would cause `isOverWorkspace` to use incorrect bounds, leading to blocks being created outside the visible workspace or the hover transition failing to fire.

For example, if the user scrolls the page down by 200px after the bounds are first cached, the cached `top`/`bottom` values are 200px off from reality. A user dragging a block over what they see as the workspace would be evaluated against the wrong rectangle.

**Suggestion:**
```
Add a `scroll` event listener (with `{ capture: true }` to catch all scroll events on ancestors) in addition to the `resize` listener:

```typescript
const handleInvalidateBounds = () => {
  wsBoundsRef.current = null;
};

document.addEventListener('scroll', handleInvalidateBounds, { capture: true, passive: true });
window.addEventListener('resize', handleInvalidateBounds);

// cleanup:
document.removeEventListener('scroll', handleInvalidateBounds, { capture: true });
window.removeEventListener('resize', handleInvalidateBounds);
```
```

File: imagelab-frontend/src/hooks/useBlocklyWorkspace.ts [hunk 2]
```diff
@@ -14,6 +14,14 @@ import {
 } from "./workspacePersistence";
 
 const SAVE_DEBOUNCE_MS = 500;
+
+const SNAP_RADIUS = 48;
+const CONNECTING_SNAP_RADIUS = 68;
+
+// Apply global Blockly configuration once at module load
```
ivantha: **[HIGH]** Global Blockly config mutated at module load time

Setting `Blockly.config.snapRadius` and `Blockly.config.connectingSnapRadius` at module scope (outside any function or hook) is a global side effect that fires the moment the module is imported — before any React component mounts. This has several consequences:

1. **Test pollution**: Any test file that imports this module will silently mutate the Blockly global config for all subsequent tests, regardless of whether those tests mock Blockly.
2. **Multiple workspace instances**: If any other part of the app imports this module and creates a Blockly workspace with different intended snap settings, those settings will be silently overridden.
3. **SSR / non-browser environments**: Module-level DOM/global mutations can cause errors in server-side or test-runner environments that don't have the Blockly config object.

To make it worse, the exact same values are set again inside `initWorkspace` (lines 42–43), making the module-level assignment completely redundant.

**Suggestion:**
```
Remove the module-level `Blockly.config` mutation entirely (lines 17–21). Keep only the instance inside `initWorkspace` where it is safe and scoped:

```typescript
// Remove these 4 lines:
Blockly.config.snapRadius = SNAP_RADIUS;
Blockly.config.connectingSnapRadius = CONNECTING_SNAP_RADIUS;

// Keep only the ones inside initWorkspace (already there):
Blockly.config.snapRadius = SNAP_RADIUS;
Blockly.config.connectingSnapRadius = CONNECTING_SNAP_RADIUS;
```

Alternatively, pass snap settings through Blockly's `inject` options if the API supports it, to keep the config scoped to the workspace instance.
```

File: imagelab-frontend/src/hooks/useSidebarDrag.ts [hunk 3]
```diff
@@ -4,26 +4,92 @@ import { SINGLETON_BLOCK_TYPES, isSingletonBlockPresent } from "../utils/blockLi
 
 const DRAG_THRESHOLD = 5;
 
+function createSyntheticPointerEvent(e: MouseEvent): PointerEvent {
+  return new PointerEvent("pointermove", {
+    clientX: e.clientX,
+    clientY: e.clientY,
+    bubbles: true,
+    pointerId: 1,
```
ivantha: **[MEDIUM]** Hardcoded `pointerId: 1` may conflict with real active pointer events

The synthetic `PointerEvent` always uses `pointerId: 1`. On a multi-touch or stylus device, pointer ID 1 may already be in use by a real pointer event. Blockly's drag internals use pointer IDs to track and route events; injecting a synthetic event with an ID that collides with a real active pointer could cause the drag system to misroute or consume real input events.

Even on a standard mouse (which typically uses pointer ID 1), there is a risk that Blockly will try to call `setPointerCapture(1)` on a DOM element and conflict with the browser's native capture of the actual mouse pointer.

Additionally, using `pointermove` as the event type for `startDrag` and `endDrag` calls is semantically incorrect — those operations are logically `pointerdown` and `pointerup` equivalents.

**Suggestion:**
```
Use a distinct, large pointer ID that is unlikely to clash with real input:

```typescript
const SYNTHETIC_POINTER_ID = 99999;

function createSyntheticPointerEvent(
  type: 'pointerdown' | 'pointermove' | 'pointerup',
  e: MouseEvent
): PointerEvent {
  return new PointerEvent(type, {
    clientX: e.clientX,
    clientY: e.clientY,
    bubbles: true,
    pointerId: SYNTHETIC_POINTER_ID,
    pointerType: 'mouse',
    isPrimary: false, // not a real primary pointer
  });
}
```
Then call `createSyntheticPointerEvent('pointerdown', e)` for `startDrag`, `'pointermove'` for `drag`, and `'pointerup'` for `endDrag`.
```

File: imagelab-frontend/src/hooks/useSidebarDrag.ts [hunk 4]
```diff
@@ -32,78 +98,151 @@ export function useSidebarDrag({ type, label, workspace, previewDataUrl }: UseSi
...
       }
 
+      if (isDragging.current && workspace) {
+        removeGhost();
+
+        if (blocklyBlockRef.current && blocklyDragActive.current) {
+          try {
+            const pointerEvent = createSyntheticPointerEvent(e);
+            blocklyBlockRef.current.endDrag(pointerEvent);
+          } catch (err) {
```
ivantha: **[MEDIUM]** Potential orphaned Blockly block when both `endDrag` and `dispose` fail

In `onMouseUp`, when `endDrag` throws an exception, the fallback `catch` block attempts to dispose the block. However, the `dispose` call is wrapped in its own `try/catch` that silently swallows errors (`catch { /* ignore */ }`). If `dispose` also throws, the block remains live in the workspace — leaked and not tracked by any ref (since `abortBlocklyDrag(true)` is called in `finally`, which sets `blocklyBlockRef.current = null` regardless).

After `abortBlocklyDrag(true)`, the ref is cleared, meaning there is no longer any handle to the orphaned block. It will be visible in the workspace with no way to programmatically remove it short of clearing the entire workspace.

```typescript
} catch (err) {
  // If this dispose() throws too, the block is leaked:
  try { blocklyBlockRef.current.dispose(false); } catch { /* ignore */ }
} finally {
  abortBlocklyDrag(true); // ref cleared — if dispose failed, block is now untracked
}
```

**Suggestion:**
```
Track whether dispose succeeded, and only call `abortBlocklyDrag(true)` if cleanup was complete:

```typescript
catch (err) {
  console.error('[useSidebarDrag] endDrag failure:', err);
  let disposed = false;
  if (blocklyBlockRef.current) {
    try {
      blocklyBlockRef.current.dispose(false);
      disposed = true;
    } catch (disposeErr) {
      console.error('[useSidebarDrag] dispose also failed:', disposeErr);
    }
  }
  // Pass success=false if block wasn't cleaned up so abortBlocklyDrag logs a warning
  abortBlocklyDrag(disposed);
  return; // skip the finally abortBlocklyDrag
} finally {
  // Only runs if no catch was hit
  abortBlocklyDrag(true);
}
```
```

PR #149, opened 2026-03-05
fix(sobel): convert float64 output to uint8 to fix pipeline encoding failure

### LINKED ISSUE #148: [Bug] Sobel/Scharr operators fail with default settings due to float64 output
**Describe the bug**
Using the **Sobel Derivative** or **Scharr Derivative** block with default parameters causes the pipeline to fail.
The backend maps `ddepth=0` to `cv2.CV_64F`, producing a `float64` image, which cannot be encoded by `cv2.imencode`.

---

**To Reproduce**

1. Upload an image using **Read Image**
2. Add a **Sobel Derivative** or **Scharr Derivative** block
3. Leave default parameters (`ddepth = 0`)
4. Run the pipeline
5. Error appears: `Failed to encode result: Could not encode image as png`

---

**Expected behavior**
The operator should return a valid output image and the pipeline should complete successfully with default settings.

---

**Screenshots**
Not applicable.

---

**Environment:**

* OS: Ubuntu / Linux
* Browser: Chrome
* Node.js version: 18.x
* Python version: 3.12

---

**Additional context**
The issue occurs because the Sobel/Scharr result is returned as `float64`. Converting the result to a displayable format (e.g., `uint8`) before encoding may resolve the issue.

Jayant-kernel:
## Description

`SobelDerivative` and `ScharrDerivative` operators failed with default parameters because the backend mapped `ddepth=0` to `cv2.CV_64F`, producing a float64 image that `cv2.imencode` cannot encode. This caused every pipeline run with default Sobel/Scharr settings to return `"Failed to encode result: Could not encode image as png"`.

Fix: always compute in `CV_64F` for gradient accuracy, then convert to uint8 via `np.uint8(np.absolute(result))` — consistent with how `Laplacian` already handles its float output. Frontend default corrected from `0` to `-1` and tooltip updated.

Fixes #148

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

Manually verified:
- Sobel Horizontal / Vertical / Both with default settings → pipeline completes, edge image displayed
- Scharr Horizontal / Vertical with default settings → pipeline completes, edge image displayed
- Output confirmed uint8, encodable as PNG

N/A

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/tests/operators/sobel_derivatives/test_scharr_derivative.py
+++ b/imagelab-backend/tests/operators/sobel_derivatives/test_scharr_derivative.py
@@ -1,0 +1,41 @@
+import cv2
+import numpy as np
+import pytest
+
+from app.operators.sobel_derivatives.scharr_derivative import ScharrDerivative
+
+
+def _sharp_edge_image() -> np.ndarray:
+    """100x100 grayscale image with a vertical edge at column 50."""
+    img = np.zeros((100, 100), dtype=np.uint8)
+    img[:, 50:] = 255
+    return img
+
+
+def test_scharr_horizontal_returns_uint8() -> None:
+    result = ScharrDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    assert result.dtype == np.uint8
+
+
+def test_scharr_vertical_returns_uint8() -> None:
+    result = ScharrDerivative({"type": "VERTICAL"}).compute(_sharp_edge_image())
+    assert result.dtype == np.uint8
+
+
+def test_scharr_output_is_png_encodable() -> None:
+    """Regression test: float64 output from cv2.Scharr used to crash cv2.imencode."""
+    result = ScharrDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    success, _ = cv2.imencode(".png", result)
+    assert success, "Scharr output must be PNG-encodable"
+
+
+def test_scharr_no_overflow_on_sharp_edge() -> None:
+    """convertScaleAbs must saturate correctly — strong edges should be near 255."""
+    result = ScharrDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    assert result.max() > 200, f"Expected strong edge response, got max={result.max()}"
+
+
+@pytest.mark.parametrize("direction", ["HORIZONTAL", "VERTICAL"])
+def test_scharr_output_shape_matches_input(direction: str) -> None:
+    result = ScharrDerivative({"type": direction}).compute(_sharp_edge_image())
+    assert result.shape == _sharp_edge_image().shape
--- a/imagelab-backend/tests/operators/sobel_derivatives/test_sobel_derivative.py
+++ b/imagelab-backend/tests/operators/sobel_derivatives/test_sobel_derivative.py
@@ -1,0 +1,56 @@
+import cv2
+import numpy as np
+import pytest
+
+from app.operators.sobel_derivatives.sobel_derivative import SobelDerivative
+
+
+def _sharp_edge_image() -> np.ndarray:
+    """100x100 grayscale image with a vertical edge at column 50."""
+    img = np.zeros((100, 100), dtype=np.uint8)
+    img[:, 50:] = 255
+    return img
+
+
+def test_sobel_horizontal_returns_uint8() -> None:
+    result = SobelDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    assert result.dtype == np.uint8
+
+
+def test_sobel_vertical_returns_uint8() -> None:
+    result = SobelDerivative({"type": "VERTICAL"}).compute(_sharp_edge_image())
+    assert result.dtype == np.uint8
+
+
+def test_sobel_both_returns_uint8() -> None:
+    result = SobelDerivative({"type": "BOTH"}).compute(_sharp_edge_image())
+    assert result.dtype == np.uint8
+
+
+def test_sobel_output_is_png_encodable() -> None:
+    result = SobelDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    success, _ = cv2.imencode(".png", result)
+    assert success, "Sobel output must be PNG-encodable"
+
+
+def test_sobel_no_overflow_on_sharp_edge() -> None:
+    """convertScaleAbs must produce a strong (near-255) response at the edge, not a wrapped low value."""
+    result = SobelDerivative({"type": "HORIZONTAL"}).compute(_sharp_edge_image())
+    assert result.max() > 200, f"Expected strong edge response, got max={result.max()}"
+
+
+def test_sobel_both_uses_l2_magnitude() -> None:
+    """BOTH direction should use np.hypot (L2 norm), not arithmetic mean."""
+    img = np.zeros((50, 50), dtype=np.uint8)
+    img[10:40, 10:40] = 255  # square produces edges in both x and y
+    result = SobelDerivative({"type": "BOTH"}).compute(img)
+    assert result.dtype == np.uint8
+    # L2 magnitude of a diagonal edge is stronger than either axis alone
+    h_result = SobelDerivative({"type": "HORIZONTAL"}).compute(img)
+    assert result.max() >= h_result.max()
+
+
+@pytest.mark.parametrize("direction", ["HORIZONTAL", "VERTICAL", "BOTH"])
+def test_sobel_output_shape_matches_input(direction: str) -> None:
+    result = SobelDerivative({"type": direction}).compute(_sharp_edge_image())
+    assert result.shape == _sharp_edge_image().shape
--- a/imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts
@@ -1,7 +1,7 @@
 export const sobelDerivativesBlocks = [
   {
     type: "sobelderivatives_soblederivate",
-    message0: "Apply %1 sobel derivative with %2 depth",
+    message0: "Apply %1 sobel derivative",
     args0: [
       {
         type: "field_dropdown",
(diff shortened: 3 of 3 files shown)


ivantha:
## ◎ Argus Review

This PR correctly identifies and patches the root cause — float64 output from Sobel/Scharr crashing `cv2.imencode` — and the fix gets pipelines unblocked. However, the chosen conversion method `np.uint8(np.absolute(result))` introduces a silent integer overflow bug that will produce visually incorrect output for images with sharp edges, replacing one defect with another. Additionally, the `ddepth` UI field is now dead code that silently discards user input, which is misleading UX. The fix lacks automated regression tests, which is risky for a bug fix on a core image operator.

### Findings (10)

#### 🔴 Silent integer overflow in np.uint8 cast will corrupt pixel values
**Severity:** critical | **Category:** bug | **File:** `imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py:15`

The conversion `np.uint8(np.absolute(result))` uses NumPy's **modular (wrapping) arithmetic** when casting float64 to uint8. Any absolute gradient value exceeding 255 wraps around silently:

```python
>>> np.uint8(300.0)   # 300 % 256 = 44
44
>>> np.uint8(512.0)   # 512 % 256 = 0  
0
>>> np.uint8(256.0)   # wraps to 0
0
```

Sobel/Scharr gradients computed in `CV_64F` on typical uint8 input images can easily produce values far above 255 — a single-pixel step from 0→255 through a 3×3 Scharr kernel yields a raw gradient value proportional to `255 * kernel_weight`. Wrapped-around pixels appear as dark artifacts instead of bright edges, producing a visually incorrect and misleading result.

The PR description says it follows the `Laplacian` pattern — if that operator has the same conversion, it carries the same latent bug.

**Suggestion:**
```
Use `cv2.convertScaleAbs(result)` — the OpenCV-recommended path for converting float gradient images back to uint8. It computes `saturate_cast<uint8>(|x| * alpha + beta)` with correct saturation (clipping at 255, not wrapping):

```python
# Before:
return np.uint8(np.absolute(result))

# After:
return cv2.convertScaleAbs(result)
```

Alternatively, explicit clipping before casting:
```python
return np.clip(np.absolute(result), 0, 255).astype(np.uint8)
```
The same fix should be applied to `sobel_derivative.py` and audited in the Laplacian operator.
```

---

#### 🔴 Same silent integer overflow in Sobel's np.uint8 cast
**Severity:** critical | **Category:** bug | **File:** `imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py:19`

Identical to the Scharr issue: `np.uint8(np.absolute(result))` will silently wrap gradient values above 255, producing incorrect pixel intensities for sharp edges. The `BOTH` direction case is additionally affected through `cv2.addWeighted`, where the pre-absolute-value arithmetic mean of `sobel_x` and `sobel_y` can still produce large-magnitude float values that overflow on cast.

For reference, a typical strong edge on a uint8 image (0→255 step) produces a Sobel gradient in the range `[−1020, +1020]` with a 3×3 kernel. `np.uint8(1020.0)` = `1020 % 256` = `252`, which looks almost correct by coincidence but is not. At other amplitudes the error is severe.

**Suggestion:**
```
```python
# In all three branches of the direction check, replace:
return np.uint8(np.absolute(result))

# With:
return cv2.convertScaleAbs(result)
```

For the `BOTH` direction, the magnitude computation itself should also be corrected (see separate finding).
```

---

#### 🟠 Incorrect gradient magnitude for 'BOTH' direction — arithmetic mean instead of L2 norm
**Severity:** high | **Category:** logic | **File:** `imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py:16`

The combined gradient in the `BOTH` case uses `cv2.addWeighted(sobel_x, 0.5, sobel_y, 0.5, 0)`, which computes:

$$M = 0.5 \cdot G_x + 0.5 \cdot G_y$$

This is the **arithmetic mean of signed gradients**, not the edge magnitude. Problems:

1. **Sign cancellation**: An edge at 45° produces a large positive $G_x$ and large positive $G_y$; an edge at 135° produces large positive $G_x$ and large *negative* $G_y$. The arithmetic mean partially cancels the 135° edge, making it appear weaker or invisible.
2. **Mathematically wrong**: Edge magnitude is $\sqrt{G_x^2 + G_y^2}$ (L2 norm) or $|G_x| + |G_y|$ (L1 approximation). Neither equals $|0.5 G_x + 0.5 G_y|$.
3. **This pre-exists the PR** but the PR's refactor is an ideal opportunity to fix it.

This is a logic error that causes diagonal edges to be systematically underrepresented in the `BOTH` output.

**Suggestion:**
```
```python
# Option 1: Proper L2 norm (most accurate, slight perf cost)
sobel_x = cv2.Sobel(image, cv2.CV_64F, 1, 0)
sobel_y = cv2.Sobel(image, cv2.CV_64F, 0, 1)
magnitude = np.sqrt(sobel_x**2 + sobel_y**2)
result = magnitude

# Option 2: L1 approximation (fast, good enough for display)
result = np.absolute(sobel_x) + np.absolute(sobel_y)

# Then in either case:
return cv2.convertScaleAbs(result)
```
```

---

#### 🟠 ddepth UI field is dead code — backend silently ignores it
**Severity:** high | **Category:** maintainability | **File:** `imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts:15`

The frontend still renders a `ddepth` number input that users can interact with, but the backend now completely ignores it — `ddepth` is no longer read from `self.params` in either `SobelDerivative` or `ScharrDerivative`. Any value a user sets in the UI is silently discarded.

The updated tooltip acknowledges this with "The 'depth' field is kept for compatibility" but this creates several problems:

- **Misleading UX**: Users who adjust `ddepth` expecting it to change behavior will see no effect and may be confused.
- **False API contract**: The frontend advertises a parameter that the backend doesn't honor.
- **Silent parameter drop**: There is no warning, no error, just silent discard. This is an antipattern that makes debugging confusing.
- **Min/max range is now semantically meaningless** since the value is ignored anyway.

**Suggestion:**
```
**Option A (preferred)**: Remove the `ddepth` field from both block definitions entirely. Since the backend hardcodes `CV_64F`, the field has no function.

```typescript
// Remove this line from both sobelderivatives_sobelderivate and sobelderivatives_scharrderivate:
{ type: "field_number", name: "ddepth", value: -1, min: -1, max: 6 },
```

**Option B**: If future configurability is planned, at minimum add a visual indicator (e.g., `enabled: false`) and tooltip note that the field is currently non-functional, or raise a backend warning log when a non-default value is passed but ignored.
```

---

#### 🟡 ddepth parameter is read from params in neither operator — breaking existing user configurations
**Severity:** medium | **Category:** maintainability | **File:** `imagelab-backend/app/operators/sobel_derivatives/scharr_derivative.py:7`

The original operators supported configurable `ddepth` (e.g., users could select `CV_16S = 3` for signed 16-bit output before further processing, or `-1` to match source depth). This is now completely removed with no deprecation path.

Any saved pipeline configurations that had a non-zero `ddepth` set will silently produce different output than expected, with no error message. This is a **silent breaking change** masquerading as a bug fix.

The PR description frames this as removing a broken feature (the 0→CV_64F mapping), but it actually removes the entire parameter's backend functionality, which was not entirely broken — only the default was.

**Suggestion:**
```
If the intent is to always use `CV_64F` internally, the backend should at least log a warning when a non-default `ddepth` is passed:

```python
ddepth = self.params.get("ddepth", -1)
if ddepth not in (-1, None):
    import logging
    logging.warning(
        "ScharrDerivative: 'ddepth' parameter is ignored; "
        "operator always computes in CV_64F and returns uint8."
    )
```

Longer term, consider removing the frontend field (Option A from the previous finding) and documenting the behavioral change in a changelog or migration note.
```

---

#### 🟡 No automated tests added for a regression-prone bug fix
**Severity:** medium | **Category:** testing | **File:** `imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py:7`

This PR fixes a bug that affected the default behavior of two operators and caused pipeline-wide failures. The PR checklist explicitly notes "Tests pass locally" is unchecked and "New tests added" is unchecked.

Without automated tests:
- The specific overflow bug introduced in this PR (silent uint8 wrapping) would not be caught.
- A future refactor could reintroduce the original float64 encoding failure.
- The `BOTH` direction gradient magnitude error is similarly undetectable.

This is a regression risk on core operators used in every edge detection pipeline.

**Suggestion:**
```
Add at minimum the following test cases to the operator test suite:

```python
import numpy as np
import cv2
from app.operators.sobel_derivatives.sobel_derivative import SobelDerivative
from app.operators.sobel_derivatives.scharr_derivative import ScharrDerivative

def make_test_image():
    # Sharp edge — will produce high gradient values to expose overflow
    img = np.zeros((100, 100), dtype=np.uint8)
    img[:, 50:] = 255
    return img

def test_sobel_returns_uint8_horizontal():
    result = SobelDerivative({"type": "HORIZONTAL"}).compute(make_test_image())
    assert result.dtype == np.uint8, f"Expected uint8, got {result.dtype}"

def test_sobel_returns_uint8_both():
    result = SobelDerivative({"type": "BOTH"}).compute(make_test_image())

def test_scharr_output_encodable_as_png():
    result = ScharrDerivative({"type": "HORIZONTAL"}).compute(make_test_image())
    success, buf = cv2.imencode('.png', result)
    assert success, "Output must be PNG-encodable"

def test_sobel_no_overflow_on_sharp_edge():
    # A sharp 0→255 edge should produce a bright (near-255) response, not a wrapped low value
    result = SobelDerivative({"type": "HORIZONTAL"}).compute(make_test_image())
    assert result.max() > 200, f"Expected strong edge response, got max={result.max()}"
```
```

---

#### 🔵 Scharr tooltip incorrectly describes it as a 'second order' derivative
**Severity:** low | **Category:** documentation | **File:** `imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts:42`

The Scharr tooltip retains the phrase "Detect edges using Scharr derivative **(second order)**" from the original code. The Scharr operator is a **first-order** derivative operator — it computes first-order spatial gradients (like Sobel), but with optimized kernel weights for better rotational symmetry. Second-order operators are things like the Laplacian ($\nabla^2 f$).

This is a pre-existing documentation error, not introduced by this PR, but a good opportunity to correct.

**Suggestion:**
```
```typescript
// Change:
"Detect edges using Scharr derivative (second order) - Applies the Scharr operator..."

// To:
"Detect edges using Scharr derivative (first order) - Applies the Scharr operator..."
```
```

---

#### 🔵 Sobel tooltip contradicts the actual operator behavior
**Severity:** low | **Category:** documentation | **File:** `imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts:22`

The new Sobel tooltip says: *"The 'depth' field is kept for compatibility but the operator always computes in CV\_64F for accuracy."*

This statement implies that `depth` can still be configured ("kept for compatibility") but in reality the backend completely ignores the field. The phrasing is ambiguous — users may interpret "kept for compatibility" as "the field still works but defaults to CV_64F" rather than the reality that it is a non-functional UI element.

**Suggestion:**
```
Be explicit about the field being non-functional:
```typescript
"...The 'depth' field is shown for UI compatibility only and has no effect — the operator always computes in CV_64F internally and returns a uint8 result for display."
```
Or better, remove the field entirely (see the ddepth dead-code finding).
```

---

#### ℹ️ Positive: Refactored to store intermediate result — improves readability
**Severity:** info | **Category:** style | **File:** `imagelab-backend/app/operators/sobel_derivatives/sobel_derivative.py:11`

The change from inline `return cv2.Sobel(...)` to storing the result in `result` and returning at the end is a clean improvement. It eliminates repeated conversion logic, makes the single return point obvious, and makes future modifications (e.g., switching to `cv2.convertScaleAbs`) straightforward without touching multiple branches. This is good practice.

**Suggestion:**
```
No change needed — just note this as a positive pattern worth keeping.
```

---

#### ℹ️ Frontend min/max range change is semantically sensible but moot given dead-code status
**Severity:** info | **Category:** logic | **File:** `imagelab-frontend/src/blocks/definitions/sobel-derivatives.blocks.ts:15`

Changing `min: -10, max: 10` to `min: -1, max: 6` aligns with actual OpenCV `cv2.CV_*` depth constants (`-1` = same as source, `0`=CV_8U through `6`=CV_64F). The original range of -10 to 10 allowed nonsensical values. This is a correct improvement if the field were functional. However, since the backend ignores `ddepth` entirely, the range is academically correct but practically irrelevant.

**Suggestion:**
```
Retain the corrected range if the field is kept, but this is moot if the field is removed as recommended in the ddepth dead-code finding.
```

---

PR #152, opened 2026-03-09
feat(filtering): add Gabor filter operator for texture detection

### LINKED ISSUE #151: Feature - Add Gabor Filter operator

ImageLab's Filtering category covers general-purpose filters (bilateral, sharpen, box, morphological, Canny) but has no operator for texture-based analysis. Users who want to detect or highlight repeating patterns, such as fabric weave, wood grain, brick mortar lines, or fingerprint ridges have no tool available without implementing a custom filter from scratch.

Add a Gabor Filter operator to the Filtering category. A Gabor filter is a Gaussian kernel modulated by a sinusoidal wave, making it selectively sensitive to edges and textures at a specific orientation and spatial frequency. It is a standard tool in classical computer vision for texture detection and feature extraction.The operator should expose five parameters via the Blockly block:

kernelSize — size of the filter window (odd, capped at 31)
sigma — width of the Gaussian envelope
theta — orientation in degrees (0–180), where 0° detects horizontal patterns and 90° detects vertical
lambda — wavelength of the sinusoidal factor, controlling texture frequency scale
gamma — spatial aspect ratio controlling filter elongation

xxivani:
## Description

Adds a Gabor Filter operator to the Filtering category for texture detection and orientation-selective edge analysis. Uses `cv2.getGaborKernel` + `cv2.filter2D` on grayscale input with `cv2.NORM_MINMAX` response normalisation. Exposes five parameters: kernel size, sigma, theta (degrees), lambda, and gamma. 

Fixes #151 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Manual testing performed in ImageLab with a brick wall image.

- [ ] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Screenshots
garbor filter

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/filtering/gabor_filter.py
+++ b/imagelab-backend/app/operators/filtering/gabor_filter.py
@@ -10,19 +10,23 @@ MAX_PROCESSING_DIM = 800
 class GaborFilter(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         kernel_size = int(self.params.get("kernelSize", 21))
-        sigma = float(self.params.get("sigma", 5.0))
+        sigma = max(0.1, float(self.params.get("sigma", 5.0)))
         theta = float(self.params.get("theta", 0.0))
-        lambda_ = float(self.params.get("lambda_", 10.0))
-        gamma = float(self.params.get("gamma", 0.5))
+        lambda_ = max(1.0, float(self.params.get("lambda", 10.0)))
+        gamma = max(0.01, float(self.params.get("gamma", 0.5)))
 
-        if kernel_size % 2 == 0:
-            kernel_size += 1
         kernel_size = max(1, min(kernel_size, MAX_KERNEL_SIZE))
+        if kernel_size % 2 == 0:
+            kernel_size = max(1, kernel_size - 1)
 
         if len(image.shape) == 3 and image.shape[2] == 4:
             image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
 
         original_h, original_w = image.shape[:2]
+
+        if original_w == 0 or original_h == 0:
+            raise ValueError(f"GaborFilter received an image with invalid dimensions: {original_w}x{original_h}")
+
         scale = min(MAX_PROCESSING_DIM / original_w, MAX_PROCESSING_DIM / original_h, 1.0)
         if scale < 1.0:
             small_w = max(1, int(original_w * scale))


ivantha:
## ◎ Argus Review

This PR adds a well-structured Gabor filter operator that follows existing project conventions and uses correct OpenCV primitives. The core algorithm (`cv2.getGaborKernel` + `cv2.filter2D` with float32 precision and `NORM_MINMAX` response normalisation) is correctly implemented, and the performance safeguard via downscaling uses the right interpolation modes. However, the backend performs no independent parameter validation for `sigma` and `lambda_`, which can cause OpenCV to produce NaN/Inf values or crash if those parameters reach zero. Additionally, the PR checklist claims new tests were added, but no test file appears anywhere in the diff — this gap is significant for a new operator with several branching paths. Minor UX inconsistencies between frontend constraints and backend limits also need attention.

### Findings (13)

#### 🟠 Test files claimed in checklist are absent from the diff
**Severity:** high | **Category:** testing | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:1`

The PR checklist marks **'New tests added'** as completed, but no test file appears anywhere in the diff. `GaborFilter.compute()` has multiple non-trivial code paths that are untested:

- BGRA → BGR auto-conversion (line 22–23)
- Grayscale (2D) input handling (line 44)
- Even `kernel_size` auto-increment (lines 18–19)
- Oversized-image downscale → filter → upscale path (lines 27–50)
- Default parameter fallback values

If the test file exists in the repository but was omitted from the diff, please surface it. If it does not exist, this is a blocking gap before merge.

**Suggestion:**
```
Create `imagelab-backend/tests/operators/filtering/test_gabor_filter.py` covering at minimum:
```python
import numpy as np
import pytest
from app.operators.filtering.gabor_filter import GaborFilter

def make_op(params):
    op = GaborFilter()
    op.params = params
    return op

def test_bgr_input_returns_bgr():
    img = np.random.randint(0, 255, (100, 100, 3), dtype=np.uint8)
    result = make_op({}).compute(img)
    assert result.shape == (100, 100, 3)

def test_bgra_input_is_converted():
    img = np.random.randint(0, 255, (100, 100, 4), dtype=np.uint8)
    result = make_op({}).compute(img)  # must not raise
    assert result.shape[2] == 3

def test_grayscale_input():
    img = np.random.randint(0, 255, (100, 100), dtype=np.uint8)
    result = make_op({}).compute(img)
    assert result.ndim == 3  # always returns BGR

def test_even_kernel_size_is_incremented():
    img = np.zeros((50, 50, 3), dtype=np.uint8)
    make_op({"kernelSize": 4}).compute(img)  # must not raise

def test_oversized_image_downscaled_then_upscaled():
    img = np.zeros((1000, 1200, 3), dtype=np.uint8)
    result = make_op({}).compute(img)
    assert result.shape == (1000, 1200, 3)
```
```

---

#### 🟠 No server-side lower-bound validation for sigma and lambda — can produce NaN/Inf or crash
**Severity:** high | **Category:** bug | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:13`

The frontend enforces `min: 0.1` on `sigma` and `min: 1.0` on `lambda_`, but the backend applies **zero independent validation**. Parameters sourced outside the UI (direct API calls, automated pipelines, scripted tests, or future batch-processing features) can deliver `sigma=0` or `lambda_=0`:

- **`sigma=0`**: The Gaussian envelope formula contains `exp(−x²/(2σ²))`. When σ=0, this is a division by zero; OpenCV fills the kernel with `NaN`/`Inf`, which silently propagates through `cv2.filter2D` and corrupts the entire output array without raising an exception.
- **`lambda_=0`**: The sinusoidal carrier `cos(2πx/λ)` is undefined at λ=0. OpenCV's behaviour is platform-dependent and may silently produce garbage or raise a `cv2.error`.
- **`gamma=0`**: Collapses the elliptical Gaussian to a line, also degenerate.

The frontend `min` attributes are a UX hint, not a security boundary.

**Suggestion:**
```
Add explicit server-side guards immediately after parameter parsing:
```python
kernel_size = int(self.params.get("kernelSize", 21))
sigma     = max(0.1,  float(self.params.get("sigma",   5.0)))
theta     =           float(self.params.get("theta",   0.0))
lambda_   = max(1.0,  float(self.params.get("lambda_", 10.0)))
gamma     = max(0.01, float(self.params.get("gamma",   0.5)))
```
Or raise a descriptive `ValueError` for invalid inputs to surface the problem clearly to callers.
```

---

#### 🟡 Odd-enforcement applied before clamping — brittle if MAX_KERNEL_SIZE ever changes to an even number
**Severity:** medium | **Category:** logic | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:18`

The current sequence:
```python
if kernel_size % 2 == 0:
    kernel_size += 1          # e.g. 30 → 31, 32 → 33
kernel_size = max(1, min(kernel_size, MAX_KERNEL_SIZE))  # 33 → 31 ✓
```
This works today **only** because `MAX_KERNEL_SIZE = 31` is odd. If it is ever changed to an even value (e.g., 32), an input of `32` would become `33`, then be clamped back to `32` (even) — violating the odd-number invariant and silently passing an even kernel size to `cv2.getGaborKernel`. While OpenCV does accept even kernel sizes, the silent violation of the stated constraint is a latent defect.

**Suggestion:**
```
Clamp first, then enforce the odd constraint so the logic is independent of `MAX_KERNEL_SIZE`'s parity:
```python
kernel_size = max(1, min(kernel_size, MAX_KERNEL_SIZE))
if kernel_size % 2 == 0:
    kernel_size = max(1, kernel_size - 1)  # round down to nearest odd
```
```

---

#### 🟡 kernelSize field_number missing `max` constraint — frontend/backend limit silently mismatched
**Severity:** medium | **Category:** maintainability | **File:** `imagelab-frontend/src/blocks/definitions/filtering.blocks.ts:121`

The `kernelSize` field has `min: 1` but no `max`. A user can type `999` in the UI; the backend silently clamps it to `31`. This creates a misleading experience where an unreachable value appears accepted. More importantly, `MAX_KERNEL_SIZE = 31` is defined only in the Python file — the frontend has no awareness of it. If the constant changes, the frontend gives no corresponding feedback to the user.

**Suggestion:**
```
```typescript
{ type: "field_number", name: "kernelSize", value: 21, min: 1, max: 31, precision: 1 },
```
Also consider adding a note in the tooltip that kernel size is always rounded up to the nearest odd number (e.g. entering 20 applies a 21-pixel kernel).
```

---

#### 🟡 Silent downscale → filter → upscale degrades output quality without user feedback
**Severity:** medium | **Category:** performance | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:25`

Images larger than 800×800 are silently downscaled before filtering and upscaled after. This performance safeguard has meaningful implications:

1. **Loss of fine texture detail**: Gabor filters are specifically designed to capture texture at a given spatial frequency. A 4K image filtered at ~800px loses sub-pixel frequency content — the operator's core use case.
2. **Lambda scale shift**: The effective spatial wavelength changes proportionally with the downscale factor. A user calibrating `lambda` against a downscaled preview will see different textural responses at different image resolutions.
3. **No user feedback**: There is nothing in the API response, logs, or UI to indicate that the image was processed at reduced resolution.

This is a defensible trade-off but should be documented and ideally configurable.

**Suggestion:**
```
At minimum, add an explanatory comment:
```python
# Performance safeguard: filter on a ≤800px working copy to keep latency reasonable.
# NOTE: This reduces texture-detail accuracy for large images. For high-fidelity
# analysis, increase MAX_PROCESSING_DIM or remove the downscaling path.
scale = min(MAX_PROCESSING_DIM / original_w, MAX_PROCESSING_DIM / original_h, 1.0)
```
Longer-term, consider exposing `MAX_PROCESSING_DIM` as a config value, or emitting a warning flag in the operator's response metadata.
```

---

#### 🟡 Using `"lambda_"` as a dictionary string key is non-idiomatic and leaks Python conventions into the public API
**Severity:** medium | **Category:** maintainability | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:15`

`lambda_` with a trailing underscore is the PEP 8 convention for a Python *variable name* that would otherwise shadow the `lambda` keyword. However, **dictionary string keys are not Python identifiers** — `"lambda"` is a perfectly legal and idiomatic string key. By using `"lambda_"` as the key, the Python-specific workaround becomes a **public API detail** that leaks into:
- The frontend Blockly parameter name (`name: "lambda_"`)
- Any external API calls or serialised pipeline configs
- Future SDK consumers or integrations

This is unnecessary and surprising to non-Python developers interacting with the API.

**Suggestion:**
```
Use `"lambda"` as the canonical parameter key in both layers:
```python
# gabor_filter.py
lambda_ = float(self.params.get("lambda", 10.0))
```
```typescript
// filtering.blocks.ts
{ type: "field_number", name: "lambda", value: 10.0, min: 1.0, precision: 0.5 },
```
The local Python variable can still be `lambda_` to avoid shadowing the keyword.
```

---

#### 🔵 Theta `max: 180` includes a degenerate value — 180° produces identical output to 0°
**Severity:** low | **Category:** logic | **File:** `imagelab-frontend/src/blocks/definitions/filtering.blocks.ts:125`

Real-valued Gabor filters have 180° rotational symmetry (they are even functions of orientation), meaning `theta = 180°` produces an output **physically identical** to `theta = 0°`. The valid non-redundant range is `[0°, 180°)` — exclusive at 180°.

The tooltip says `'Theta' (0–180°)` which implies inclusion of 180°, potentially confusing users who try to sweep the full orientation space and are surprised to return to horizontal-edge detection at the boundary.

**Suggestion:**
```
```typescript
{ type: "field_number", name: "theta", value: 0, min: 0, max: 179, precision: 1 },
```
Update the tooltip to state the range is `0°–179°` and clarify that 0° and 180° are equivalent.
```

---

#### 🔵 GaborFilter class and compute() method lack docstrings
**Severity:** low | **Category:** documentation | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:10`

There is no class-level or method-level docstring. Gabor filtering is mathematically involved and several non-obvious decisions are made here: `psi=0` (cosine phase), `ktype=cv2.CV_32F`, taking `np.abs()` of the bipolar response, and the downscaling heuristic. Future maintainers will need to reverse-engineer these choices without guidance.

**Suggestion:**
```
```python
class GaborFilter(BaseOperator):
    """
    Applies a Gabor filter for texture detection and orientation-selective edge analysis.

    A Gabor filter is a Gaussian-windowed sinusoid, sensitive to edges and textures
    at a specific orientation (theta) and spatial frequency (lambda). The filter is
    applied to the grayscale version of the input; the magnitude of the bipolar
    response is normalised to [0, 255] and returned as a BGR image.

    Large images (> MAX_PROCESSING_DIM) are downscaled before filtering to limit
    latency; accuracy may be reduced for fine-detail texture analysis.
    """

        """
        Apply Gabor filter to the input image.

        Args:
            image: Input BGR, BGRA, or grayscale uint8 NumPy array.

        Returns:
            BGR uint8 NumPy array with Gabor magnitude response normalised to [0, 255].
        """
```
```

---

#### 🔵 Phase offset psi=0 is hardcoded and unexplained
**Severity:** low | **Category:** maintainability | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:40`

`psi` (the sinusoidal carrier phase) is hardcoded to `0`, selecting a **cosine-phase** Gabor filter. This is the most common default and produces symmetric edge-sensitive responses. However, `psi=π/2` produces a sine-phase filter that is more sensitive to lines (bright bars on dark backgrounds) rather than edges. Without a comment, future maintainers cannot tell whether `psi=0` was a deliberate choice or an oversight, and may hesitate to modify it.

**Suggestion:**
```
```python
kernel = cv2.getGaborKernel(
    (kernel_size, kernel_size),
    sigma,
    np.deg2rad(theta),
    lambda_,
    gamma,
    psi=0,           # cosine phase → edge-sensitive; use np.pi/2 for line-sensitive
    ktype=cv2.CV_32F,
)
```
Alternatively, expose `psi` as an advanced (collapsed/expert) UI parameter.
```

---

#### 🔵 Tooltip string is ~430 characters — too long for a UI tooltip widget
**Severity:** low | **Category:** style | **File:** `imagelab-frontend/src/blocks/definitions/filtering.blocks.ts:134`

The tooltip is more than 430 characters across a single concatenated string. Typical UI tooltip implementations (including Blockly's) either truncate long tooltips, render them in a small fixed-width box causing awkward wrapping, or overflow on mobile/small-screen viewports. The valuable per-parameter descriptions belong in documentation or a contextual help panel, not a hover tooltip.

**Suggestion:**
```
Shorten to a focused one- or two-sentence summary:
```typescript
tooltip:
  "Apply a Gabor filter for orientation-selective texture detection. " +
  "Theta (0–179°) selects edge orientation; Lambda controls the texture frequency scale.",
```
Move the full parameter breakdown to the project's operator documentation or an inline help-panel component.
```

---

#### ℹ️ Grayscale input always produces BGR output — channel behaviour is inconsistently documented
**Severity:** info | **Category:** logic | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:44`

If a 2D (single-channel grayscale) image is passed in, the code correctly skips the BGRA→BGR and BGR→GRAY conversions, applies the filter, but then unconditionally runs `cv2.cvtColor(result_gray, cv2.COLOR_GRAY2BGR)`. This means the output is always 3-channel BGR, regardless of input channel count.

This may be intentional (standardised output format for downstream pipeline nodes), but it is not documented and may diverge from the behaviour of other operators in the filtering category.

**Suggestion:**
```
If BGR output is always desired for pipeline consistency, add a comment:
```python
# Always return BGR for consistent downstream pipeline behaviour,
# even when input was grayscale.
result_small = cv2.cvtColor(result_gray, cv2.COLOR_GRAY2BGR)
```
If the intent is to mirror the input channel count, add a conditional:
```python
if len(image.shape) == 3:
    result_small = cv2.cvtColor(result_gray, cv2.COLOR_GRAY2BGR)
else:
    result_small = result_gray
```
```

---

#### ℹ️ Potential ZeroDivisionError on zero-dimension images
**Severity:** info | **Category:** bug | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:26`

```python
scale = min(MAX_PROCESSING_DIM / original_w, MAX_PROCESSING_DIM / original_h, 1.0)
```
If `original_w` or `original_h` is `0` (e.g., a synthetically constructed or malformed array), this raises an unhandled `ZeroDivisionError` with no contextual message. This is an unlikely edge case with real image data but trivially triggerable in tests or via adversarial API inputs.

**Suggestion:**
```
```python
if original_w == 0 or original_h == 0:
    raise ValueError(
        f"GaborFilter received an image with invalid dimensions: {original_w}×{original_h}"
    )
```
```

---

#### ℹ️ Positive: several implementation details reflect strong OpenCV/signal-processing practice
**Severity:** info | **Category:** style | **File:** `imagelab-backend/app/operators/filtering/gabor_filter.py:1`

The following choices are worth highlighting as good patterns to preserve and replicate in future operators:

- **`INTER_AREA` for downsampling, `INTER_LINEAR` for upsampling** (lines 30, 50): the textbook-correct interpolation pair that minimises aliasing on the way down and avoids nearest-neighbour artefacts on the way up.
- **`ktype=cv2.CV_32F` and `cv2.CV_32F` ddepth** (lines 41, 45): avoids quantisation error during convolution, critical for correct Gabor response magnitude.
- **`np.abs(filtered)`** (line 46): correctly handles the bipolar (positive and negative) Gabor response before normalising to uint8 — taking the magnitude is standard practice.
- **BGRA stripping before grayscale conversion** (lines 22–23): prevents incorrect alpha-premultiplication artefacts in `cvtColor`.
- **Registry key follows existing naming convention** (`filtering_gaborfilter`): consistent with `filtering_erosion`, `filtering_morphological`, etc.

**Suggestion:**
```
No change needed — these are patterns to preserve and reuse in future operator implementations.
```

---

PR #161, opened 2026-03-10
fix: remove invalid OpenCV constants from DistanceTransform and cleanup frontend block

### LINKED ISSUE #160: [Bug] DistanceTransform uses incorrect OpenCV constants for DIST_LABEL_PIXEL and DIST_MASK_3
**Describe the bug**
The **Distance Transform** block exposes `DISTLABEL_PIXEL` and `DISTMASK_3` as distance type options, but both behave the same as `DISTL2`. Selecting either option produces identical output, meaning the dropdown options have no effect.

The issue appears to come from incorrect constant mapping in the backend where these options are mapped to `cv2.DIST_L2`.

---

**To Reproduce**

1. Upload any image.
2. Add a **Distance Transform** block.
3. In the **distance type dropdown**, select `DISTLABEL_PIXEL` or `DISTMASK_3`.
4. Run the pipeline.
5. Compare the output with `DISTL2`.

Result: Output remains identical to `DISTL2`.

---

**Expected behavior**
Selecting different distance types should produce different distance transform results.

The dropdown should only contain valid `cv2.distanceTransform` distance types such as:

* `DIST_C`
* `DIST_L1`
* `DIST_L2`

Currently exposed options like `DISTLABEL_PIXEL` and `DISTMASK_3` belong to different OpenCV enum families and are not valid distance type parameters.

---

**Environment:**

* OS: Windows 11
* Browser: Chrome
* Node.js version: 18+
* Python version: 3.10+

---

**Additional context**

Backend file:

`imagelab-backend/app/operators/transformation/distance_transform.py`

Current constant mapping:

```python
DISTANCE_TYPES = {
    "DIST_C": cv2.DIST_C,
    "DIST_L1": cv2.DIST_L1,
    "DIST_L2": cv2.DIST_L2,
    "DIST_LABEL_PIXEL": cv2.DIST_L2,
    "DIST_MASK_3": cv2.DIST_L2,
}
```

`cv2.DIST_LABEL_PIXEL` is part of the **DistanceLabelType enum**, and `cv2.DIST_MASK_3` belongs to the **DistanceMaskSize enum**. Neither should be used as a `distanceType` argument in `cv2.distanceTransform`.

Jayant-kernel:
## Description

Removes `DIST_LABEL_PIXEL` and `DIST_MASK_3` from the `DISTANCE_TYPES` dict in `distance_transform.py`. These are **not** valid distance type constants in OpenCV — they belong to the `DistanceLabelType` and `DistanceMaskSize` enum families respectively. Both were incorrectly mapped to `cv2.DIST_L2`, causing silently wrong output whenever a user selected either option from the dropdown.

Also removes the unused `ddepth` field from the frontend block definition since the backend never reads it for this operator.

Fixes #160

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass (`uv run pytest`)
- [ ] New tests added
- [x] Manual testing

Manual verification:
- Selected `DIST_C` → output correctly differs from `DIST_L2` (Chebyshev distance)
- Selected `DIST_L1` → output correctly differs from `DIST_L2` (Manhattan distance)
- Selected `DIST_L2` → standard Euclidean distance output unchanged
- Frontend block dropdown now shows only the 3 valid distance type options

N/A — silent wrong output bug, no visual diff to show.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/transformation/distance_transform.py
+++ b/imagelab-backend/app/operators/transformation/distance_transform.py
@@ -7,8 +7,6 @@ DISTANCE_TYPES = {
     "DIST_C": cv2.DIST_C,
     "DIST_L1": cv2.DIST_L1,
     "DIST_L2": cv2.DIST_L2,
-    "DIST_LABEL_PIXEL": cv2.DIST_L2,
-    "DIST_MASK_3": cv2.DIST_L2,
 }
 
 
--- a/imagelab-frontend/src/blocks/definitions/transformation.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/transformation.blocks.ts
@@ -1,20 +1,17 @@
 export const transformationBlocks = [
   {
     type: "transformation_distance",
-    message0: "Apply %1 distance with %2 depth",
+    message0: "Apply %1 distance transform",
     args0: [
       {
         type: "field_dropdown",
         name: "type",
         options: [
-          ["DISTC", "DIST_C"],
-          ["DISTL1", "DIST_L1"],
-          ["DISTL2", "DIST_L2"],
-          ["DISTLABEL_PIXEL", "DIST_LABEL_PIXEL"],
-          ["DISTMASK_3", "DIST_MASK_3"],
+          ["DIST_C", "DIST_C"],
+          ["DIST_L1", "DIST_L1"],
+          ["DIST_L2", "DIST_L2"],
         ],
       },
-      { type: "field_number", name: "ddepth", value: 0, min: -10, max: 10 },
     ],
     previousStatement: null,
     nextStatement: null,


imagelab-backend/app/operators/transformation/distance_transform.py
```diff
@@ -7,8 +7,6 @@
     "DIST_C": cv2.DIST_C,
     "DIST_L1": cv2.DIST_L1,
     "DIST_L2": cv2.DIST_L2,
-    "DIST_LABEL_PIXEL": cv2.DIST_L2,
-    "DIST_MASK_3": cv2.DIST_L2,
 }
```
ivantha: **[INFO]** ✅ Correct removal of cross-enum constants — silent wrong-output bug eliminated

The removed entries were genuinely incorrect. `cv2.DIST_LABEL_PIXEL` belongs to `cv2.DistanceLabelTypes` (used as the `labelType` parameter in `cv2.distanceTransformWithLabels`), and `cv2.DIST_MASK_3` belongs to `cv2.DistanceMaskSize` (used as the `maskSize` parameter). Neither is a valid `distanceType` argument to `cv2.distanceTransform`. Mapping both to `cv2.DIST_L2` was doubly wrong:

- **`DIST_LABEL_PIXEL`** has numeric value `1`, which equals `cv2.DIST_L1`. So passing it would accidentally compute L1 (Manhattan) distance — a completely different metric.
- **`DIST_MASK_3`** has numeric value `3`, which equals `cv2.DIST_C` (Chebyshev). So passing it would accidentally compute Chebyshev distance, also wrong.

In both cases, the silent aliasing to `cv2.DIST_L2` (value `2`) in the dict overrides what the integer value would have produced if passed raw — making this particularly hard to debug.

The fix is correct and the analysis in the PR description is sound.

**Suggestion:**
```
No change needed — this is a positive observation confirming the fix is correct.
```

ivantha: **[MEDIUM]** No regression test added to guard against re-introduction of invalid constants

The PR removes two silently-wrong entries from `DISTANCE_TYPES` but does not add a test to ensure they cannot be re-introduced. Without a test, a future contributor could add new constants to this dict (e.g., confusing another `DistanceMaskSize` or `DistanceLabelTypes` member) and cause the same class of silent bug.

Additionally, a test verifying that each remaining key (`DIST_C`, `DIST_L1`, `DIST_L2`) actually produces **distinct** output for a non-trivial input image would catch future cross-aliasing bugs (e.g., if two keys were accidentally mapped to the same `cv2` constant).

The PR checklist notes 'New tests added: ☐', acknowledging this gap.

**Suggestion:**
```
Add a unit test such as:

```python
# test_distance_transform.py
import cv2
import numpy as np
from app.operators.transformation.distance_transform import DISTANCE_TYPES

def test_distance_types_keys_are_valid_cv2_distance_types():
    valid_cv2_distance_types = {
        cv2.DIST_L1, cv2.DIST_L2, cv2.DIST_C,
        cv2.DIST_L12, cv2.DIST_FAIR, cv2.DIST_WELSCH, cv2.DIST_HUBER
    }
    for name, val in DISTANCE_TYPES.items():
        assert val in valid_cv2_distance_types, (
            f"'{name}' maps to {val}, which is not a valid cv2 DistanceTypes member"
        )

def test_distance_types_produce_distinct_outputs():
    # A non-trivial binary image so metrics actually differ
    img = np.zeros((20, 20), dtype=np.uint8)
    img[5:15, 5:15] = 255
    results = {
        name: cv2.distanceTransform(img, dist_type, 3).sum()
        for name, dist_type in DISTANCE_TYPES.items()
    }
    # DIST_C, DIST_L1, DIST_L2 must all produce different sums
    assert len(set(results.values())) == len(results), (
        f"Some distance types produce identical output: {results}"
    )
```
```

PR #167, opened 2026-03-11
fix(transformation): correct Laplacian ddepth handling and expose ksi…

### LINKED ISSUE #166: [Bug] - Laplacian operator silently remaps ddepth=0 to CV_64F; tooltip and docs describe incorrect behaviour
**Describe the bug**

The Laplacian operator has three related issues that together make it behave differently from what the UI describes. In laplacian.py, ddepth=0 (the block's default) is silently remapped to cv2.CV_64F so the output depth is always 64-bit float regardless of what the user sets. The block tooltip states "a value of 0 means the output will have the same depth as the source image" which is incorrect because 1 is the OpenCV convention for same-depth passthrough, not 0. Additionally, operatorDocs.ts documents KSize as the Laplacian's parameter (which is a real and useful cv2.Laplacian argument controlling aperture size) but neither the block nor the backend exposes ksize at all and it is always hardcoded to the OpenCV default of 1.

**To Reproduce**

1. Open ImageLab and load any image
2. Add a Read Image block followed by a Laplacian block
3. Set ddepth to any value including 0 and run the pipeline
4. Observe that the backend always processes with cv2.CV_64F regardless of the value entered
5. Read the tooltip and note it states 0 means same-depth passthrough which is factually incorrect per OpenCV docs

**Expected behavior**

ddepth=-1 should be treated as same depth as source matching OpenCV convention. ddepth=0 should behave as CV_8U which is what 0 actually maps to in OpenCV. The block default should be -1 so out-of-the-box behaviour produces safe same-depth output. The tooltip should accurately describe what -1 does. The ksize parameter should be exposed in the block and backend so users can control aperture size with valid odd values 1, 3, 5, and 7. operatorDocs.ts should document ksize not ddepth as the primary parameter.

**Screenshots**

**Environment:**
- Browser: [e.g. Chrome 120]
- Python version: [e.g. 3.12]

**Additional context**

xxivani:
## Description
Fixes three related issues in the Laplacian operator: a silent `ddepth=0 → CV_64F` remap in the backend, an incorrect tooltip claiming `ddepth=0` means same-depth passthrough when the correct OpenCV value is `-1`, and a missing `ksize` parameter that was documented in `operatorDocs.ts` but never exposed in the block or backend.

Fixes #166 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Manually verified that a Laplacian block with default settings (`ddepth=-1`, `ksize=1`) produces correct edge output on a test image. Verified that `ksize=3` and `ksize=7` produce visibly broader edge responses. Confirmed the previous behaviour where all ddepth values silently produced CV_64F output no longer occurs.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/transformation/laplacian.py
+++ b/imagelab-backend/app/operators/transformation/laplacian.py
@@ -3,18 +3,22 @@ import numpy as np
 
 from app.operators.base import BaseOperator
 
-_VALID_KSIZE = {1, 3, 5, 7}
+_VALID_KSIZE: frozenset[int] = frozenset({1, 3, 5, 7})
+_VALID_DDEPTH: frozenset[int] = frozenset({-1, cv2.CV_8U, cv2.CV_16U, cv2.CV_16S, cv2.CV_32F, cv2.CV_64F})
 
 
 class Laplacian(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         ddepth = int(self.params.get("ddepth", -1))
-        if ddepth == -1:
-            ddepth = -1
+        if ddepth not in _VALID_DDEPTH:
+            raise ValueError(f"Invalid ddepth={ddepth}; must be one of {sorted(_VALID_DDEPTH)}")
+
+        if ddepth == -1 and np.issubdtype(image.dtype, np.integer):
+            ddepth = cv2.CV_64F
 
         ksize = int(self.params.get("ksize", 1))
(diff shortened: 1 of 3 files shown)


xxivani:
Hi @ivantha!
All findings have been addressed.

PR #172, opened 2026-03-11
feat: interactive learning mode — step previews, histogram analysis and pipeline persistence (#39)

Jayant-kernel:
## Description

Implements Features 1, 2, and 3 from issue #39 (per-step intermediate previews, image analysis panel with histogram, and pipeline persistence) with full test coverage and all correctness issues from the existing PR #98 review addressed.

**Why this supersedes PR #98:**

PR #98 implements Features 1 & 2 but has several correctness issues flagged by @ivantha that remain unresolved:
- Float-image histogram bug: `cv2.calcHist` with range `[0, 256]` produces meaningless counts for float images — fixed here by normalising to uint8 before computing histograms
- Architectural inversion: `utils` imported from `models` — fixed here by returning a plain dict from `compute_image_stats()`, keeping utils dependency-free
- BGRA alpha channel included in histograms — fixed here by capping at `min(channels, 3)`
- Missing `type="button"` on toolbar buttons — fixed here
- No payload size control — fixed here (JPEG thumbnails max 200 px, list endpoint returns summaries only, no pipeline_json bloat)
- Syntax error in `StepPreviewList.tsx` preventing compilation — not present here
- Zero tests — this PR adds 31 tests

Additionally, this PR implements **Feature 3 (Pipeline Persistence)** which PR #98 does not cover at all.

Fixes #39

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

**New tests (31 total):**
- `tests/utils/test_image_stats.py` — histogram correctness for uint8/float/BGRA/uniform images, float regression for PR #98 bug
- `tests/services/test_pipeline_executor_intermediates.py` — flag off/on, step count, operator types, thumbnail non-empty, stats correctness, final image still returned
- `tests/routers/test_saved_pipelines.py` — full CRUD via `TestClient` + in-memory SQLite (`get_db` override); covers 404s, invalid JSON 422, delete+refetch

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/filtering/gabor_filter.py
+++ b/imagelab-backend/app/operators/filtering/gabor_filter.py
@@ -1,0 +1,55 @@
+import cv2
+import numpy as np
+
+from app.operators.base import BaseOperator
+
+MAX_KERNEL_SIZE = 31
+MAX_PROCESSING_DIM = 800
+
+
+class GaborFilter(BaseOperator):
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        kernel_size = int(self.params.get("kernelSize", 21))
+        sigma = max(0.1, float(self.params.get("sigma", 5.0)))
+        theta = float(self.params.get("theta", 0.0))
+        lambda_ = max(1.0, float(self.params.get("lambda", 10.0)))
+        gamma = max(0.01, float(self.params.get("gamma", 0.5)))
+
+        kernel_size = max(1, min(kernel_size, MAX_KERNEL_SIZE))
+        if kernel_size % 2 == 0:
+            kernel_size = max(1, kernel_size - 1)
+
+        if len(image.shape) == 3 and image.shape[2] == 4:
+            image = cv2.cvtColor(image, cv2.COLOR_BGRA2BGR)
+
+        original_h, original_w = image.shape[:2]
+
+        if original_w == 0 or original_h == 0:
+            raise ValueError(f"GaborFilter received an image with invalid dimensions: {original_w}x{original_h}")
+
+        scale = min(MAX_PROCESSING_DIM / original_w, MAX_PROCESSING_DIM / original_h, 1.0)
+        if scale < 1.0:
+            small_w = max(1, int(original_w * scale))
+            small_h = max(1, int(original_h * scale))
+            working = cv2.resize(image, (small_w, small_h), interpolation=cv2.INTER_AREA)
+        else:
+            working = image
+
+        kernel = cv2.getGaborKernel(
+            (kernel_size, kernel_size),
+            sigma,
+            np.deg2rad(theta),
+            lambda_,
+            gamma,
+            psi=0,
+            ktype=cv2.CV_32F,
+        )
+
+        gray = cv2.cvtColor(working, cv2.COLOR_BGR2GRAY) if len(working.shape) == 3 else working
+        filtered = cv2.filter2D(gray.astype(np.float32), cv2.CV_32F, kernel)
+        result_gray = cv2.normalize(np.abs(filtered), None, 0, 255, cv2.NORM_MINMAX).astype(np.uint8)
+        result_small = cv2.cvtColor(result_gray, cv2.COLOR_GRAY2BGR)
+
+        if scale < 1.0:
+            return cv2.resize(result_small, (original_w, original_h), interpolation=cv2.INTER_LINEAR)
+        return result_small
--- a/imagelab-backend/app/operators/filtering/morphological.py
+++ b/imagelab-backend/app/operators/filtering/morphological.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import cv2
 import numpy as np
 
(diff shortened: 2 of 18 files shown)


imagelab-backend/app/models/saved_pipeline.py
```diff
@@ -0,0 +1,55 @@
...
+    id: int | None = Field(default=None, primary_key=True)
+    name: str = Field(max_length=200, index=True)
+    description: str = Field(default="", max_length=1000)
+    pipeline_json: str  # JSON-serialised list[PipelineStep]
+    workspace_state: str = Field(default="")  # JSON-serialised Blockly workspace state
+    share_token: str = Field(
+        default_factory=lambda: secrets.token_urlsafe(12),
+        unique=True,
+        index=True,
+    )
```
ivantha: **[HIGH]** Share token entropy may be insufficient for public sharing

`secrets.token_urlsafe(12)` generates 12 bytes → 16 base64url characters → ~96 bits of entropy. While this is technically sufficient against brute-force, the tokens are stored in a **non-rate-limited** public endpoint (`GET /pipelines/share/{token}`). An adversary can enumerate tokens at API speed. There is no rate limiting, no authentication on the list endpoint, and tokens are returned in the list response.

Furthermore, `GET /pipelines` (list endpoint) returns `share_token` for every pipeline in the `SavedPipelineSummary` response model. This means anyone with read access to the list endpoint can trivially obtain all share tokens — rendering the sharing mechanism meaningless as a privacy control.

This is only a concern if pipeline data is considered sensitive, but it is worth flagging architecturally.

**Suggestion:**
```
1. **Remove `share_token` from `SavedPipelineSummary`** if the list endpoint is unauthenticated, or add authentication to the list endpoint.
2. Consider adding rate limiting to `GET /pipelines/share/{token}` (e.g. via `slowapi`).
3. If per-user isolation is intended, add a user ownership column and filter list/delete by the authenticated user.

At minimum, document the intended security model in a comment.
```

imagelab-frontend/src/components/SavedPipelinesModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+        }
+      } catch {
+        Blockly.serialization.workspaces.load(snapshot, workspace);
+        setLoadError("Failed to restore workspace state");
+        return;
+      }
+      workspace.getBlocksByType(READ_IMAGE_BLOCK_TYPE, false).forEach((block) => {
+        block.getField(FILENAME_LABEL_FIELD)?.setValue("No image");
+      });
+      onClose();
```
ivantha: **[HIGH]** Rollback on workspace load failure uses stale snapshot taken before clear

In `handleLoad`, the code:
1. Takes a snapshot: `const snapshot = Blockly.serialization.workspaces.save(workspace)`
2. Clears the workspace: `workspace.clear()`
3. Tries to load the saved state — if this throws, it attempts to restore from snapshot

However, after `workspace.clear()`, the workspace may be in a partially mutated state (event listeners, undo stack, etc.). Restoring via `Blockly.serialization.workspaces.load(snapshot, workspace)` into an already-cleared workspace **may not fully restore the previous state**, particularly the undo history and any registered change listeners.

This is a UX correctness issue: the user is told "Failed to restore workspace state" but the workspace they see may differ from what they had before clicking Load.

**Suggestion:**
```
Consider deferring the `workspace.clear()` until after the new state is verified as loadable. One approach is to parse and validate `saved.workspace_state` (JSON parse + basic schema check) *before* clearing, and only clear+load if validation passes:

```typescript
const handleLoad = async (id: number) => {
  const saved = await getPipeline(id);
  if (!saved.workspace_state) { onClose(); return; }
  let parsed: unknown;
  try {
    parsed = JSON.parse(saved.workspace_state);
  } catch {
    setLoadError("Saved workspace state is corrupt");
    return;
  }
  // Only now commit to clearing
  if (existingBlocks.length > 0) {
    if (!window.confirm("...")) return;
  }
  workspace.clear();
  Blockly.serialization.workspaces.load(parsed as ..., workspace);
  onClose();
};
```
```

File: imagelab-backend/app/services/pipeline_executor.py [hunk 3]
```diff
@@ -65,14 +106,29 @@ def execute_pipeline(request: PipelineRequest) -> PipelineResponse:
...
+        # Intermediate capture is isolated from operator execution: a stats/thumbnail
+        # failure must never turn a successful pipeline step into an error response.
+        if collect:
+            try:
+                raw_stats = compute_image_stats(image)
+                intermediates.append(
+                    StepResult(
+                        step=i + 1,
+                        operator_type=step.type,
+                        thumbnail=_make_thumbnail(image),
```
ivantha: **[MEDIUM]** Intermediates collected for NOOP steps

The `NOOP_TYPES` set is defined and used to skip operator execution for certain step types (e.g. `basic_readimage`, `basic_writeimage`). However, the intermediate capture block runs unconditionally for **every** step index `i` after the operator is applied — including NOOP steps.

Looking at the broader executor logic (not fully shown), NOOP steps presumably `continue` or return early without modifying `image`. If a NOOP step is included, it will generate an intermediate that reflects the *previous* step's output, which is misleading (it attributes the previous image state to the NOOP operator type).

This depends on the full executor logic, but the PR description mentions NOOP_TYPES exist — the intermediate capture should mirror the same guard.

**Suggestion:**
```
Add a NOOP guard around the intermediate capture block:

```python
if collect and step.type not in NOOP_TYPES:
    try:
        raw_stats = compute_image_stats(image)
        intermediates.append(
            StepResult(
                step=i + 1,
                operator_type=step.type,
                thumbnail=_make_thumbnail(image),
                stats=ImageStats(**raw_stats),
            )
        )
    except Exception:
        logger.warning("Failed to capture intermediate for step %d (%s)", i + 1, step.type, exc_info=True)
```
```

File: imagelab-backend/app/models/pipeline.py [hunk 4]
```diff
@@ -23,10 +24,29 @@ class PipelineTimings(BaseModel):
...
+class ImageStats(BaseModel):
+    width: int
+    height: int
+    channels: int
+    dtype: str
+    min_val: float
+    max_val: float
+    mean_val: float
+    histograms: list[list[int]]  # one 256-bucket list per channel (max 3)
+
```
ivantha: **[LOW]** `histograms` field comment says "max 3" but grayscale images return 1

The `ImageStats` model documents:

```python
histograms: list[list[int]]  # one 256-bucket list per channel (max 3)
```

The "max 3" is correct (alpha excluded), but the comment doesn't mention the minimum is 1 for grayscale images. Frontend consumers need to handle both `len == 1` and `len == 3` cases, which `HistogramCanvas` does correctly (via `isGray` check), but the API contract could be clearer.

**Suggestion:**
```
Update the comment:

```python
histograms: list[list[int]]  # 1 list for grayscale, 3 lists for colour (B/G/R); alpha excluded
```

Also consider adding the same detail to the TypeScript `ImageStats` interface in `pipeline.ts`.
```

Jayant-kernel:
@ivantha 
 I understand this task is reserved for the selected GSoC student. I built this as a way to demonstrate my understanding of the codebase and what the feature could look like in practice .
I  will submit my draft  GSoC proposal for this project today and would be great if you take a look . Looking forward to the selection .

PR #179, opened 2026-03-12
Fix ColorToBinary crash on grayscale input

### LINKED ISSUE #178: [Bug] ColorToBinary crashes on grayscale input
**Difficulty:** Easy
**Area:** backend

## Description
`ColorToBinary.compute` branches only on the 4-channel case, so every other input falls through to `cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)` at `color_to_binary.py:22`. A single-channel `(H, W)` array therefore reaches `cvtColor` with a conversion code that requires 3 or 4 channels and raises a raw `cv2.error` about a bad number of channels. The graph endpoint hides this: `OPERATOR_PORT_SPECS` declares `imageconvertions_colortobinary` as accepting only 3 or 4 channel input, and `_validate_graph` rejects the connection before anything executes. That contract is not enforced on the flat pipeline path, where `create_execution` skips `compile_graph` entirely whenever `request.graph` is `None` and hands the step list straight to `execute_pipeline`, so the crash is still reachable.

## Steps to Reproduce
1. `POST /v1/pipeline/executions` with a flat `pipeline` list and no `graph` field, so the graph validation is skipped.
2. Use the two steps `[{"type": "imageconvertions_grayimage"}, {"type": "imageconvertions_colortobinary"}]` against any colour image.
3. `GrayImage` returns a 2-D `(H, W)` array, which is passed to `ColorToBinary`.

## Expected vs Actual
- **Expected:** an already single-channel image is used as the grayscale source and thresholded directly, or is rejected with a readable operator-level error.
- **Actual:** `cv2.cvtColor` raises `cv2.error` for a bad number of channels, and the step result carries the raw OpenCV message.

## Acceptance Criteria
- [ ] `ColorToBinary.compute` accepts `(H, W)` and `(H, W, 1)` inputs and thresholds them without calling `cv2.cvtColor`
- [ ] 3-channel and 4-channel behaviour is unchanged
- [ ] An unsupported channel count raises a `ValueError` with a readable message instead of a raw `cv2.error`
- [ ] The guard follows the shape already used by `PrewittOperator`, so the two conversion paths stay consistent
- [ ] Regression test in `imagelab-backend/tests/test_conversion_operators.py` running `ColorToBinary` over a single-channel array
- [ ] `uv run pytest -v` passes

## Resources
- `imagelab-backend/app/operators/conversions/color_to_binary.py:19` - the 4-channel branch, and the unguarded `else` on line 22 that assumes 3 channels
- `imagelab-backend/app/services/graph_engine.py:18` - `OPERATOR_PORT_SPECS`, where `imageconvertions_colortobinary` is declared as `[3, 4]` channels
- `imagelab-backend/app/services/graph_engine.py:243` - `_validate_graph`, which enforces that contract but runs only via `compile_graph`
- `imagelab-backend/app/routers/pipeline.py:38` - `compile_graph` is called only when `request.graph is not None`, leaving the flat path unvalidated
- `imagelab-backend/app/operators/sobel_derivatives/prewitt_operator.py:38` - the 1/3/4 channel handling pattern to copy
- [cv2.cvtColor colour conversions](https://docs.opencv.org/4.x/d8/d01/group__imgproc__color__conversions.html)

<details>
<summary>Original report by @Jayant-kernel (2026-03-12)</summary>

**Describe the bug**
`ColorToBinary` crashes when the input image is already grayscale (single-channel).

**To Reproduce**
1. Upload an image that is grayscale (1-channel), or pass output from `GrayImage`.
2. Add `ColorToBinary` in the pipeline.
3. Run processing.
4. Backend throws an exception and the pipeline fails.

**Expected behavior**
`ColorToBinary` should handle grayscale input safely (either return valid binary output directly or convert appropriately) without crashing.

**Screenshots**
N/A

**Environment:**
- OS: Ubuntu 22.04
- Browser: Chrome
- Node.js version: 18.x
- Python version: 3.12

**Additional context**
This is similar in pattern to other conversion operators that crash when channel assumptions are not validated.

</details>

Jayant-kernel:
## Description

Fix `ColorToBinary` crash when the input is already grayscale (single-channel).

Changes:
- Handle grayscale input directly without `cvtColor`
- Support BGRA input via `cv2.COLOR_BGRA2GRAY`
- Keep existing BGR behavior
- Add regression tests for grayscale and BGRA inputs

Fixes #178

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added tests in `imagelab-backend/tests/test_conversion_operators.py`:
- `test_grayscale_input_does_not_crash`
- `test_bgra_input_supported`

Local execution in this environment:
- `python -m pytest imagelab-backend/tests/test_conversion_operators.py -q` failed to start because `cv2` is not installed (`ModuleNotFoundError: No module named 'cv2'`).

- [ ] Existing tests pass
- [x] New tests added
- [ ] Manual testing

N/A

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/conversions/color_to_binary.py
+++ b/imagelab-backend/app/operators/conversions/color_to_binary.py
@@ -15,11 +15,11 @@ class ColorToBinary(BaseOperator):
         threshold_type = THRESHOLD_TYPES.get(threshold_type_name, cv2.THRESH_BINARY)
         threshold_value = float(self.params.get("thresholdValue", 0))
         max_value = float(self.params.get("maxValue", 0))
-
-        if len(image.shape) == 3 and image.shape[2] == 4:
+        if image.ndim == 2:
+            gray = image
+        elif image.ndim == 3 and image.shape[2] == 4:
             gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
         else:
             gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
-
         _, dst = cv2.threshold(gray, threshold_value, max_value, threshold_type)
         return dst
--- a/imagelab-backend/tests/test_conversion_operators.py
+++ b/imagelab-backend/tests/test_conversion_operators.py
@@ -113,6 +113,19 @@ class TestColorToBinary:
         result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(color_image)
         assert result.dtype == np.uint8
 
+    def test_grayscale_input_does_not_crash(self, grayscale_image):
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(grayscale_image)
+        assert result.shape == grayscale_image.shape
+        assert result.dtype == np.uint8
+        assert set(np.unique(result)).issubset({0, 255})
+
+    def test_bgra_input_supported(self, color_image):
+        alpha = np.full(color_image.shape[:2], 255, dtype=np.uint8)
+        bgra = np.dstack([color_image, alpha])
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bgra)
+        assert result.shape == color_image.shape[:2]
+        assert result.dtype == np.uint8
+
 
 # ColorMaps
 


Jayant-kernel:
@ivantha 
review my pr

ivantha (commented):
## ◎ Argus Review

This PR correctly identifies and fixes a real crash in `ColorToBinary` when grayscale images are passed, and adds support for BGRA inputs as a proactive improvement. The core fix logic is sound for the most common cases, but the implementation has a gap: 3D single-channel images with shape `(H, W, 1)` are not handled and will still crash. The `else` branch also silently falls through to BGR conversion for any unrecognized format, which could mask incorrect usage. On the testing side, the new tests are well-structured but were never actually executed (cv2 not installed locally), and the `grayscale_image` fixture used by the new test does not appear in the diff — if it doesn't already exist in the file, the test will fail at collection time. Overall this is a worthwhile bug fix that needs a few targeted improvements before it is production-ready.

### ⚠️ Breaking Changes

- **[INFO]** No breaking changes — patch semver appropriate: **Previous behavior**: `ColorToBinary.compute()` always called `cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)`, which would raise a `cv2.error` for any non-BGR-3-channel input.

**New behavior**: The function now gracefully handles grayscale (2D), BGR (3-channel), and BGRA (4-channel) inputs. The BGR code path is identical to before.

**Who is affected**: No existing consumers are affected — the BGR path is unchanged, and the previously crashing paths now work.

**Migration path**: None required. This is a pure additive fix with no API signature changes.
- **[INFO]** No breaking changes — patch-level fix: **Previous behavior**: `ColorToBinary.compute()` unconditionally called `cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)`. Passing a grayscale (2D) or BGRA (4-channel) image raised an OpenCV assertion error.

**New behavior**: Grayscale images are passed through directly; BGRA images use `COLOR_BGRA2GRAY`; BGR images retain their existing conversion path.

**Who is affected**: Any caller that was previously relying on the crash behavior as a validation gate (extremely unlikely) would be affected. All other callers are unaffected or positively affected.

**Migration path**: None required. The public API signature (`compute(image: np.ndarray) -> np.ndarray`) is unchanged. The change is strictly additive: inputs that previously crashed now work correctly.
  > **Suggestion:** No action needed. This should be tagged as a `patch` release.
- **[INFO]** No breaking changes detected — pure additive bug fix: **Previous behavior**: `ColorToBinary.compute()` accepted only BGR images (H, W, 3). Passing grayscale (H, W) or BGRA (H, W, 4) input would crash with an OpenCV error.

**New behavior**: `compute()` now correctly handles grayscale (H, W), BGRA (H, W, 4), and the original BGR (H, W, 3) inputs.

**Who is affected**: No existing consumers are affected — the BGR code path is unchanged. The only behavioral change is that previously-crashing inputs now produce valid output.

**Migration path**: No migration required. This is a non-breaking additive fix.


### PR Description Issues

- **[MEDIUM]** PR description acknowledges unverified tests — checklist should block merge: The description honestly notes that tests could not be run locally due to a missing `cv2` dependency. However, this is listed in the body text, while the formal checklist items — "Existing tests pass" and "Tests pass locally" — are left **unchecked**. This is correct behavior from the author, but it means this PR should not be merged as-is without CI confirmation that the tests pass.

The description does not explain **why** the test environment lacked `cv2` or what steps were taken to work around this (e.g., running in Docker, deferring to CI). Reviewers should be given enough information to assess risk.
  > **Suggestion:** Update the description to include CI/CD test results or a note explaining how the fix was verified (e.g., "Verified via GitHub Actions CI run #XYZ"). If CI is the source of truth for test results, add a link to the passing pipeline run:

```
## Verification
Tests could not be run locally (cv2 not installed). CI results: [link to passing Actions run]
```
- **[INFO]** PR description is clear, well-structured, and accurately describes the changes: The PR description correctly enumerates all three code paths introduced (grayscale passthrough, BGRA conversion, existing BGR path), names both new tests, references the issue being fixed (#178), and honestly flags the test-execution gap. The "Type of Change" and "Checklist" sections are appropriately filled in. No template boilerplate or placeholder HTML comments were left in.
  > **Suggestion:** No changes needed for structure or accuracy. The only improvement would be adding a CI verification link as noted in the adjacent finding.
- **[MEDIUM]** PR admits tests were not verified — unchecked CI box is a workflow gap: The PR description explicitly states:
> `python -m pytest ... failed to start because cv2 is not installed (ModuleNotFoundError: No module named 'cv2')`

And leaves the `Existing tests pass` checkbox unchecked. This means the PR was submitted without any local or CI confirmation that the new tests (or existing tests) pass. Combined with the missing `grayscale_image` fixture found in review, there is a real possibility the test suite is broken.

This is a process concern: the contributor environment was not set up correctly, and the PR was merged without verified test execution.
  > **Suggestion:** Add a CI requirement to the contribution guidelines: all PRs must pass `pytest` in a Docker/virtual environment with `cv2` installed before review. The PR description should be updated to:
```
## How Has This Been Tested?
- [x] New tests added (`test_grayscale_input_does_not_crash`, `test_bgra_input_supported`)
- [ ] Tests verified in CI (cv2 not available in local dev environment — see #178 for CI setup)
```
Also consider adding a `requirements-test.txt` or `tox.ini` so contributors can easily spin up the correct environment.
- **[LOW]** Description omits the technical root cause of the crash: The description says the PR "Fix[es] `ColorToBinary` crash when the input is already grayscale" but does not explain *why* the crash occurs. A reader skimming the description does not understand the mechanism: `cv2.cvtColor` with `COLOR_BGR2GRAY` raises an OpenCV assertion error when the source array is not 3-channel, which is what happens when a grayscale (2D) image is passed.

Understanding the root cause helps reviewers evaluate whether the fix is complete and helps future contributors avoid re-introducing the bug.
  > **Suggestion:** Expand the description section:
```markdown
## Root Cause
`cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)` requires a 3-channel (BGR) source image. Passing a 2D grayscale ndarray or a 4-channel BGRA array causes an OpenCV assertion failure: `(-215:Assertion failed) scn == 3 || scn == 4 in function 'cvtColor'`.

## Fix
Dispatch on `image.ndim` and `image.shape[2]` to select the appropriate conversion (or bypass it entirely for already-grayscale inputs).
```
- **[HIGH]** Tests explicitly not verified to pass — PR submitted with known unvalidated state: The PR description explicitly states: *"python -m pytest ... failed to start because cv2 is not installed (ModuleNotFoundError: No module named 'cv2')"*. Both the `Existing tests pass` and `Tests pass locally` checklist items are unchecked. This means the new tests — the primary validation mechanism for this bug fix — have never been executed in any environment by the author.

This is a significant quality concern for a bug fix PR: it's possible the `grayscale_image` fixture is missing (see related finding), the assertions are wrong, or the fix itself doesn't behave as expected. The PR should not be merged until tests are verified to pass in a CI environment with `cv2` installed.
  > **Suggestion:** Before merge, confirm the CI pipeline runs and passes. Update the PR description to reflect test status:
```markdown
## How Has This Been Tested?

Added tests in `imagelab-backend/tests/test_conversion_operators.py`:
- `test_grayscale_input_does_not_crash`
- `test_bgra_input_supported`

**Note**: Tests were not runnable locally due to `cv2` not being installed in the development environment. Tests were validated via CI (link to passing run).

- [x] Existing tests pass (CI)
- [x] New tests added
- [ ] Manual testing
```
- **[LOW]** PR description omits motivation/context for the BGRA support addition: The description mentions that BGRA support was added (`Support BGRA input via cv2.COLOR_BGRA2GRAY`), but provides no context for *why* this was included as part of a crash fix for grayscale input. These are logically separate changes: fixing a crash vs. adding a new supported format. Without context, reviewers cannot determine whether BGRA support was required by the issue, was a proactive addition, or was contributed by a related bug report.
  > **Suggestion:** Add a brief rationale for the BGRA addition:
```markdown
## Changes
- **Fix crash on grayscale input**: Handle 2D arrays directly without calling `cvtColor`, 
  which previously raised an OpenCV error for single-channel images (fixes #178).
- **Add BGRA support** *(proactive improvement)*: While fixing the channel-dispatch logic, 
  extended support to 4-channel BGRA images, which are common when images include an 
  alpha channel (e.g., PNG screenshots).
- Keep existing BGR behavior unchanged.
```

imagelab-backend/tests/test_conversion_operators.py
```diff
@@ -105,6 +105,19 @@ def test_output_is_uint8(self, color_image):
         result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(color_image)
         assert result.dtype == np.uint8
 
+    def test_grayscale_input_does_not_crash(self, grayscale_image):
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(grayscale_image)
+        assert result.shape == grayscale_image.shape
+        assert result.dtype == np.uint8
+        assert set(np.unique(result)).issubset({0, 255})
+
+    def test_bgra_input_supported(self, color_image):
```
ivantha: **[HIGH]** Potentially missing `grayscale_image` fixture

The new test `test_grayscale_input_does_not_crash` depends on a `grayscale_image` pytest fixture, but this fixture does not appear anywhere in the diff. If it was not already present in the test file before this PR, pytest will raise a `fixture 'grayscale_image' not found` error and the test will never execute — silently defeating the purpose of the regression test.

Because the entire motivation for this PR is fixing a crash on grayscale input, the absence of a verified, runnable test for that path is a significant gap. The PR description also confirms that tests could not be run locally, so this cannot have been caught during authoring.

**Suggestion:**
```
Add a `grayscale_image` fixture to the test file (likely in the same conftest or class setup block where `color_image` is defined):

```python
@pytest.fixture
def grayscale_image():
    return np.random.randint(0, 256, (100, 100), dtype=np.uint8)
```

If the fixture already exists in the file, make it visible in the PR diff by including the surrounding context.
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py [hunk 2]
```diff
@@ -15,6 +15,11 @@ def compute(self, image: np.ndarray) -> np.ndarray:
         threshold_type = THRESHOLD_TYPES.get(threshold_type_name, cv2.THRESH_BINARY)
         threshold_value = float(self.params.get("thresholdValue", 0))
         max_value = float(self.params.get("maxValue", 0))
-        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            gray = image
+        elif image.ndim == 3 and image.shape[2] == 4:
+            gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
+        else:
+            gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
```
ivantha: **[MEDIUM]** Else branch silently assumes BGR for any non-2 or non-4-channel input

The revised conditional handles `ndim == 2` (grayscale) and `ndim == 3, channels == 4` (BGRA) explicitly, but all other cases fall through to `cv2.COLOR_BGR2GRAY`. This means:

- An image with `shape (H, W, 1)` — a "3-D but single-channel" array — will cause `cv2.cvtColor` to raise a cryptic OpenCV error rather than being handled gracefully.
- An image with 2 channels or 5+ channels will likewise throw an undescriptive `cv2` exception.
- An RGB image (channels == 3, but ordered R-G-B instead of B-G-R) will silently produce an incorrect grayscale conversion with no warning.

The `else` branch is effectively a catch-all that only works correctly for the BGR-3-channel case.

**Suggestion:**
```
Add an explicit guard and a descriptive error for unsupported formats:

```python
if image.ndim == 2:
    gray = image
elif image.ndim == 3 and image.shape[2] == 4:
    gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
elif image.ndim == 3 and image.shape[2] == 3:
    gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
else:
    raise ValueError(
        f"Unsupported image format: ndim={image.ndim}, "
        f"channels={image.shape[2] if image.ndim == 3 else 'N/A'}. "
        "Expected grayscale (H,W), BGR (H,W,3), or BGRA (H,W,4)."
    )
```
```

File: imagelab-backend/tests/test_conversion_operators.py [hunk 3]
```diff
@@ -105,6 +105,19 @@ def test_output_is_uint8(self, color_image):
...
+        alpha = np.full(color_image.shape[:2], 255, dtype=np.uint8)
+        bgra = np.dstack([color_image, alpha])
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bgra)
+        assert result.shape == color_image.shape[:2]
+        assert result.dtype == np.uint8
+
 
 # ColorMaps
```
ivantha: **[MEDIUM]** Tests authored without being able to run them — correctness unverified

The PR description explicitly states:
> `python -m pytest imagelab-backend/tests/test_conversion_operators.py -q` failed to start because `cv2` is not installed (`ModuleNotFoundError: No module named 'cv2'`).

Both new tests were written without ever being executed. This means bugs in the tests themselves (e.g., a missing fixture, an incorrect assertion, or an import error) cannot have been caught. Merging unverified tests introduces noise into the test suite — they may always fail in CI for unexpected reasons, or worse, silently skip/error in a way that masks the regression.

**Suggestion:**
```
Before merging, verify the tests pass in an environment with `cv2` available (e.g., via a Docker image, a CI pipeline step, or a virtual environment with `opencv-python` installed). The PR checklist item "Existing tests pass" and "Tests pass locally" should both be checked before approval. If the test environment cannot be set up locally, CI results should be shared.
```

ivantha: **[LOW]** `test_bgra_input_supported` does not assert binary output values

The grayscale test correctly asserts that the result contains only binary pixel values:
```python
assert set(np.unique(result)).issubset({0, 255})
```

However, the BGRA test omits this assertion. The primary purpose of `ColorToBinary` is to produce a binary image — verifying that property for BGRA input is just as important as for any other input format. Without it, the test only confirms that the function doesn't crash and returns the right shape/dtype, but does not validate the core semantics.

**Suggestion:**
```
Add the binary-values assertion to `test_bgra_input_supported`:

```python
def test_bgra_input_supported(self, color_image):
    alpha = np.full(color_image.shape[:2], 255, dtype=np.uint8)
    bgra = np.dstack([color_image, alpha])
    result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bgra)
    assert result.shape == color_image.shape[:2]
    assert set(np.unique(result)).issubset({0, 255})  # <-- add this
```
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py (hunk 2 above)
Comment (ivantha) [inline imagelab-backend/app/operators/conversions/color_to_binary.py]: **[LOW]** No comment or docstring explaining supported input formats

The function now silently accepts three different input formats (grayscale 2D, BGR 3-channel, BGRA 4-channel) but there is no documentation — inline comment or docstring — explaining this. Future developers looking at the function signature (`image: np.ndarray`) have no way to know which formats are valid without reading the implementation. This is especially important because the BGR assumption (rather than RGB) is a non-obvious OpenCV convention.

**Suggestion:**
```
Add a short comment block above the conditional or update the function docstring:

```python
# Supported image formats:
#   - Grayscale: ndim == 2, shape (H, W)
#   - BGR:       ndim == 3, shape (H, W, 3)  [OpenCV default]
#   - BGRA:      ndim == 3, shape (H, W, 4)
if image.ndim == 2:
    ...
```
```

File: imagelab-backend/tests/test_conversion_operators.py (hunk 3 above)
Comment (ivantha) [inline imagelab-backend/tests/test_conversion_operators.py]: **[INFO]** Positive: Tests directly validate the fixed invariants

Both new tests are well-scoped and validate the right properties:
- `test_grayscale_input_does_not_crash` confirms the crash is fixed, checks shape preservation, dtype correctness, and that output values are binary.
- `test_bgra_input_supported` correctly constructs a BGRA image using `np.dstack` (rather than relying on external test data) and validates shape/dtype.

The naming convention (`does_not_crash`, `input_supported`) is clear and communicates intent. The tests follow the pattern of existing tests in the class.

File: imagelab-backend/tests/test_conversion_operators.py [hunk 4]
```diff
@@ -105,6 +105,19 @@ def test_output_is_uint8(self, color_image):
         result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(color_image)
         assert result.dtype == np.uint8
 
+    def test_grayscale_input_does_not_crash(self, grayscale_image):
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(grayscale_image)
+        assert result.shape == grayscale_image.shape
+        assert result.dtype == np.uint8
+        assert set(np.unique(result)).issubset({0, 255})
```
ivantha: **[HIGH]** Missing `grayscale_image` fixture — tests will fail at collection time

The new test `test_grayscale_input_does_not_crash` references a `grayscale_image` pytest fixture as a parameter, but the diff does not show this fixture being defined anywhere in the test file. If it does not already exist in a `conftest.py` or earlier in this file (outside the diff hunk), pytest will raise a `fixture 'grayscale_image' not found` error and **all tests in this class** will fail to collect — including the pre-existing ones.

This is a straightforward oversight that completely blocks the new test from running.

**Suggestion:**
```
Add a `grayscale_image` fixture to `conftest.py` or to the test class/module:
```python
@pytest.fixture
def grayscale_image():
    """Returns a (100, 100) uint8 grayscale array."""
    return np.random.randint(0, 256, (100, 100), dtype=np.uint8)
```
Also verify the `color_image` fixture shape to ensure `np.dstack` in `test_bgra_input_supported` produces the expected `(H, W, 4)` array.
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py (hunk 2 above)
Comment (ivantha) [inline imagelab-backend/app/operators/conversions/color_to_binary.py]: **[MEDIUM]** Unhandled channel counts silently fall through to BGR conversion and crash

The new conditional handles `ndim == 2` and `ndim == 3 with shape[2] == 4`, then falls to `else` for everything else. This `else` branch unconditionally calls `cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)`, which **requires exactly 3 channels**. The following inputs will hit the `else` branch and crash with an OpenCV error:

- 3D arrays where `shape[2]` is not 3 or 4 (e.g., 1, 2, 5, 6 channels — common with hyperspectral or stacked-frame data)
- Arrays with `ndim > 3` (e.g., a batch dimension accidentally left in)
- A single-channel image stored as 3D with shape `(H, W, 1)`

The original code had the same problem for these cases, but the PR's stated goal is to fix crashes, so these should be addressed proactively.

**Suggestion:**
```
Add explicit channel count validation:
```python
if image.ndim == 2:
    gray = image
elif image.ndim == 3:
    channels = image.shape[2]
    if channels == 1:
        gray = image[:, :, 0]  # squeeze single-channel 3D array
    elif channels == 3:
        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
    elif channels == 4:
        gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
    else:
        raise ValueError(
            f"Unsupported number of channels: {channels}. Expected 1, 3, or 4."
        )
else:
    raise ValueError(
        f"Unsupported image shape: {image.shape}. Expected 2D or 3D array."
    )
```
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py [hunk 5]
```diff
@@ -15,6 +15,11 @@ def compute(self, image: np.ndarray) -> np.ndarray:
         threshold_type = THRESHOLD_TYPES.get(threshold_type_name, cv2.THRESH_BINARY)
         threshold_value = float(self.params.get("thresholdValue", 0))
         max_value = float(self.params.get("maxValue", 0))
-        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            gray = image
+        elif image.ndim == 3 and image.shape[2] == 4:
+            gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
```
ivantha: **[MEDIUM]** 4-channel input assumed to be BGRA — silently produces wrong output for RGBA images

The condition `image.ndim == 3 and image.shape[2] == 4` applies `cv2.COLOR_BGRA2GRAY` to **any** 4-channel image. However, 4-channel images from common sources use RGBA channel ordering:

- `PIL.Image.convert('RGBA')` → RGBA
- `matplotlib` saved PNGs → RGBA
- `imageio.imread()` on PNGs → RGBA

When an RGBA image is treated as BGRA, the blue and red channels are swapped before grayscale conversion, producing **subtly incorrect luminance values** with no error or warning. This is a silent data correctness bug.

**Suggestion:**
```
Either (a) expose the channel format as an operator parameter, or (b) document the BGRA assumption prominently and consider detecting the format from image metadata if available. At minimum, add a docstring/comment:
```python
# Assumes 4-channel images are in BGRA format (OpenCV convention).
# RGBA images (e.g., from PIL or matplotlib) must be converted first.
elif image.ndim == 3 and image.shape[2] == 4:
    gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
```
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py [hunk 6]
```diff
@@ -15,6 +15,11 @@ def compute(self, image: np.ndarray) -> np.ndarray:
         threshold_type = THRESHOLD_TYPES.get(threshold_type_name, cv2.THRESH_BINARY)
         threshold_value = float(self.params.get("thresholdValue", 0))
         max_value = float(self.params.get("maxValue", 0))
-        gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+        if image.ndim == 2:
+            gray = image
```
ivantha: **[LOW]** Grayscale branch assigns a reference, not a copy

When `image.ndim == 2`, the code sets `gray = image` (a reference alias). While `cv2.threshold` writes its output to a new array `dst` (so the immediate bug is benign), this pattern is fragile:

1. Any future code added between the assignment and `cv2.threshold` that mutates `gray` in-place will also mutate the caller's `image` array — a surprising side-effect.
2. It is inconsistent with the `cv2.cvtColor` branches, which always produce a new array.

This is a low-severity latent bug but worth fixing for defensive correctness.

**Suggestion:**
```
```python
if image.ndim == 2:
    gray = image.copy()  # avoid aliasing the caller's array
```
```

File: imagelab-backend/tests/test_conversion_operators.py [hunk 7]
```diff
@@ -105,6 +105,19 @@ def test_output_is_uint8(self, color_image):
...
+
+    def test_bgra_input_supported(self, color_image):
+        alpha = np.full(color_image.shape[:2], 255, dtype=np.uint8)
+        bgra = np.dstack([color_image, alpha])
+        result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bgra)
+        assert result.shape == color_image.shape[:2]
+        assert result.dtype == np.uint8
```
ivantha: **[LOW]** `test_bgra_input_supported` does not assert binary output values

`test_grayscale_input_does_not_crash` validates that the result contains only `{0, 255}` via `set(np.unique(result)).issubset({0, 255})`, which confirms the thresholding step produced a true binary image. The sister test `test_bgra_input_supported` only checks shape and dtype — it would pass even if the grayscale conversion computed wrong values that happen to threshold into a non-binary result (though `cv2.threshold` guarantees binary output for `THRESH_BINARY`, verifying this is still good test hygiene and mirrors the other test).

**Suggestion:**
```
Add the missing assertion for consistency:
```python
def test_bgra_input_supported(self, color_image):
    alpha = np.full(color_image.shape[:2], 255, dtype=np.uint8)
    bgra = np.dstack([color_image, alpha])
    result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bgra)
    assert result.shape == color_image.shape[:2]
    assert set(np.unique(result)).issubset({0, 255})  # add this
```
```

File: imagelab-backend/tests/test_conversion_operators.py (hunk 3 above)
Comment (ivantha) [inline imagelab-backend/tests/test_conversion_operators.py]: **[LOW]** No test covering error behavior for unsupported channel counts

The PR adds happy-path tests for the new supported formats (grayscale and BGRA) but does not add a test for an input that should be rejected — e.g., a 5-channel or 2-channel image. Without this, the regression suite does not specify or enforce what the operator's contract is for invalid inputs, making future refactors more risky.

**Suggestion:**
```
```python
def test_unsupported_channel_count_raises(self):
    bad_image = np.zeros((50, 50, 5), dtype=np.uint8)
    with pytest.raises(ValueError, match="Unsupported number of channels"):
        ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(bad_image)
```
(This becomes relevant once the unhandled-channel-count bug above is fixed.)
```

File: imagelab-backend/app/operators/conversions/color_to_binary.py (hunk 2 above)
Comment (ivantha) [inline imagelab-backend/app/operators/conversions/color_to_binary.py]: **[INFO]** Positive: clean conditional dispatch is readable and extensible

The if/elif/else structure is easy to read and clearly expresses intent. Each branch is a single statement, and the logic is easy to extend (e.g., adding an RGB branch would be a one-liner). This is a good pattern for a conversion operator that may eventually need to handle more formats.

**Suggestion:**
```
No change needed. When the unhandled channel count issue is addressed, this structure will naturally accommodate the additional `channels == 1` and error cases.
```

ivantha:

Jayant-kernel:
rebase done

ivantha:
Thanks for the contribution! This PR became not-mergeable (branch can't be rebased) after other PRs landed on main. Closing for now — if you'd still like to get this in, please rebase on the latest main and reopen. Your fix is appreciated.

PR #190, opened 2026-03-12
fix: default DrawText y to 30 to prevent text rendering above image

### LINKED ISSUE #189: [Bug] - DrawText block renders no text when using default parameters
**Describe the bug**
The DrawText block renders nothing when used with default settings. cv2.putText uses the text baseline as the y origin, so the default starting_point_y=0 places the text completely above the top edge of the image and nothing is visible.

**To Reproduce**

1. Add a Read Image block and upload any image
2. Add a Draw Text block and connect it
3. Run the pipeline without changing any DrawText parameters
4. The output image shows no text

**Expected behavior**
The default parameters should produce visible text on the image. The y coordinate default should be set to a value that places text within the image bounds.

**Screenshots**

**Environment:**
- Browser: [e.g. Chrome 120]
- Python version: [e.g. 3.12]

**Additional context**

xxivani:
## Description
Changes the default `starting_point_y` from `0` to `30` in both the `DrawText` backend operator and the frontend block definition. `cv2.putText` treats y as the text baseline origin, so a default of `0` renders text entirely above the image with nothing visible. Also updates the tooltip to explain that y is the baseline, not the top edge.

Fixes #189 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Ran the pipeline with a DrawText block at default parameters and confirmed text is visible on the output image.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist
- [x] Tests pass locally


ivantha (commented):
## ◎ Argus Review

This is a focused, well-reasoned bug fix that correctly addresses a real usability issue: `cv2.putText` uses y as the text baseline, so a default of `y=0` renders text entirely above the image frame. The fix keeps the backend and frontend defaults in sync at `30`, and the tooltip update adds genuinely useful documentation for users. The main concern is that `30` is a magic number that only works reliably at `scale=1`; at larger font scales (e.g., scale=2 or 3), the text may still be partially clipped. A more robust solution would compute the default y dynamically using `cv2.getTextSize`. Overall the PR is a clean, low-risk improvement with minor areas for hardening.

### ⚠️ Breaking Changes

- **[INFO]** Default starting_point_y changed from 0 to 30 — silent behavioral change for existing default-value pipelines: **Previous behavior**: Any pipeline using the DrawText block without explicitly setting `starting_point_y` rendered text at y=0 (text baseline above the top of the image — invisible).

**New behavior**: The same pipelines now render text at y=30 (text baseline 30 px below the top edge — visible).

**Who is affected**: Users with saved block graphs or serialized pipeline configs that omit `starting_point_y` (i.e., relied on the implicit default). Pipelines with an explicit `starting_point_y: 0` are entirely unaffected since the fallback default is only used when the key is absent.

**Migration path**: No migration is needed for most users; the change corrects a non-functional default. Users who intentionally set `starting_point_y` to `0` (to hide text programmatically as a workaround) should add an explicit `starting_point_y: 0` to their configs to preserve that behavior.
  > **Suggestion:** No code change required. Document the behavior change in the changelog or release notes as a bug fix: `fix: DrawText default y changed from 0 (invisible) to 30 (visible) — only affects pipelines that relied on the implicit default.`
- **[INFO]** Behavioral default value change — existing pipelines unaffected, new blocks change: **Previous behavior**: New `DrawText` blocks placed on the canvas defaulted to `starting_point_y = 0`, resulting in invisible text.

**New behavior**: New `DrawText` blocks default to `starting_point_y = 30`, rendering text visibly near the top of the image.

**Who is affected**: Only newly created `DrawText` blocks are affected. Existing saved pipelines that already have an explicit `starting_point_y` value (including `0`) stored in their configuration will continue to use that stored value — the backend fallback default only applies when the key is entirely absent from params, which does not occur in normal pipeline execution.

**Migration path**: No migration needed. The change is backward-compatible. Any user who intentionally used `y=0` (though the result was invisible) would need to manually update their block, but this is a degenerate case with no practical use.
  > **Suggestion:** No action required. This is a safe, non-breaking behavioral improvement.


### PR Description Issues

- **[INFO]** Missing before/after screenshots for a visual rendering bug: The PR description states "Screenshots (if applicable)" with no content. Since this is explicitly a visual rendering fix — text was invisible, now it is visible — before/after screenshots are directly applicable and would make the fix immediately verifiable by reviewers without needing to run the pipeline locally.
  > **Suggestion:** Add two annotated screenshots under the Screenshots section:
- **Before**: a pipeline output showing the DrawText block with y=0 (empty/unchanged image)
- **After**: a pipeline output showing the DrawText block with y=30 (text visible near the top)

This directly validates the claimed fix and aligns with the manual testing noted in the checklist.
- **[INFO]** PR description does not address backward compatibility of saved pipelines: Users who have saved pipelines where the DrawText block was created with the default y (i.e., they never explicitly set y) will now see those pipelines render text at y=30 rather than y=0. While this is an improvement (y=0 was broken), it is a silent behavioral change for existing saved state. The PR description does not mention this.
  > **Suggestion:** Add a brief compatibility note, for example:

> **Saved pipeline compatibility**: Pipelines where `starting_point_y` was explicitly set to a value (including `0`) are unaffected. Only pipelines that relied on the implicit default will change behavior — they will now render text at `y=30` instead of the previously invisible `y=0`.
- **[LOW]** Screenshots section left empty despite this being a visual rendering bug fix: The PR description includes a **Screenshots** section but leaves it completely blank. For a bug fix whose entire purpose is to make visual output *visible*, before/after screenshots would be the single most compelling form of evidence that the fix works correctly. The absence of screenshots makes it harder for reviewers to quickly verify the fix without running the project locally.
  > **Suggestion:** Add before/after screenshots showing:
- **Before**: A DrawText block output with default parameters where text is invisible (or the blank canvas)
- **After**: The same block output where text is clearly rendered near the top of the image

This is especially impactful given that the fix is about visual visibility.
- **[INFO]** PR description is well-structured and accurately matches the diff: The description correctly identifies:
1. The root cause (`cv2.putText` treats y as the text **baseline**, not the top edge)
2. The exact files changed (backend operator + frontend block definition)
3. The specific value changed (`0` → `30`)
4. The tooltip update
5. Links to the relevant issue (#189)

The 'Type of Change' checklist is correctly marked as a bug fix only. The testing section is honest about not adding new automated tests. The overall description quality is high.
  > **Suggestion:** No major changes needed. Adding screenshots (see separate finding) and a note about the scale-dependency of the `30` value would make this description excellent.

ivantha:
Reviewed. Fixes an invisible-text bug in the DrawText operator by computing a dynamic default y using `cv2.getTextSize` (text height + baseline) rather than hardcoding 0, which placed text entirely above the image. The frontend block default and tooltip are updated accordingly, and new backend tests verify visibility, shape preservation, and immutability. Merging via rebase.

ivantha:

xxivani:
Hi @ivantha, I've pushed the updated branch!

PR #192, opened 2026-03-12
fix: expose tx and ty translation params on AffineImage operator

### LINKED ISSUE #191: [Bug] - AffineImage transformation is hardcoded and non-configurable
**Describe the bug**

The AffineImage operator always applies a fixed translation of 50px in X and 100px in Y regardless of user input. The transformation matrix M = np.float64([[1, 0, 50], [0, 1, 100]]) is hardcoded in the backend with no parameters exposed, making the block completely non-configurable.

**To Reproduce**

1. Add a Read Image block and upload any image
2. Add an Affine Transform block and connect it
3. Run the pipeline
4. The image always shifts by exactly 50px right and 100px down with no way to change it

**Expected behavior**
The block should expose tx and ty translation parameters so the user can control how the image is shifted.

**Screenshots**

**Environment:**
- Browser: [e.g. Chrome 120]
- Python version: [e.g. 3.12]

**Additional context**

xxivani:
## Description
Exposes `tx` and `ty` translation parameters on the AffineImage operator, replacing the hardcoded `M = np.float64([[1, 0, 50], [0, 1, 100]])` matrix. The frontend block now shows two number fields for X and Y translation. The backend reads these params to construct the transformation matrix at runtime. Default values of `tx=50, ty=100` preserve the previous behaviour. `operatorDocs.ts` updated to document the new parameters.

Fixes #191 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Ran the pipeline with the AffineImage block at various `tx`/`ty` values and confirmed the image shifts correctly in both axes.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/geometric/affine_image.py
+++ b/imagelab-backend/app/operators/geometric/affine_image.py
@@ -7,5 +7,7 @@ from app.operators.base import BaseOperator
 class AffineImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         rows, cols = image.shape[:2]
-        M = np.float64([[1, 0, 50], [0, 1, 100]])
+        tx = float(self.params.get("tx", 50))
+        ty = float(self.params.get("ty", 100))
+        M = np.float64([[1, 0, tx], [0, 1, ty]])
         return cv2.warpAffine(image, M, (cols, rows))
--- a/imagelab-backend/app/operators/geometric/affine_image.py
+++ b/imagelab-backend/app/operators/geometric/affine_image.py
@@ -1,3 +1,5 @@
+import math
+
 import cv2
 import numpy as np
 
@@ -5,9 +7,23 @@ from app.operators.base import BaseOperator
 
 
 class AffineImage(BaseOperator):
+    DEFAULT_TX = 50.0
+    DEFAULT_TY = 100.0
+
     def compute(self, image: np.ndarray) -> np.ndarray:
         rows, cols = image.shape[:2]
-        tx = float(self.params.get("tx", 50))
-        ty = float(self.params.get("ty", 100))
+
+        try:
+            tx = float(self.params.get("tx", self.DEFAULT_TX))
+            ty = float(self.params.get("ty", self.DEFAULT_TY))
+        except (TypeError, ValueError):
+            tx, ty = self.DEFAULT_TX, self.DEFAULT_TY
+
+        if not math.isfinite(tx) or not math.isfinite(ty):
+            tx, ty = self.DEFAULT_TX, self.DEFAULT_TY
+
+        tx = max(-cols, min(cols, tx))
+        ty = max(-rows, min(rows, ty))
+
         M = np.float64([[1, 0, tx], [0, 1, ty]])
         return cv2.warpAffine(image, M, (cols, rows))
--- a/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
@@ -96,11 +96,15 @@ export const geometricBlocks = [
   },
   {
     type: "geometric_affineimage",
-    message0: "Apply affine transformation",
+    message0: "Apply affine transformation | translate x %1 y %2",
+    args0: [
+      { type: "field_number", name: "tx", value: 50 },
+      { type: "field_number", name: "ty", value: 100 },
+    ],
     previousStatement: null,
     nextStatement: null,
     style: "geometric_style",
     tooltip:
-      "Applies a fixed affine transformation (translate by 50, 100) - Transforms the image using a predefined affine transformation that translates the image by 50 pixels in the X direction and 100 pixels in the Y direction. This block is useful for demonstrating basic affine transformations, which can include translation, rotation, scaling, and shearing.",
+      "Applies an affine transformation to the image - Translates the image by the specified number of pixels along the X and Y axes. Positive tx shifts the image right; positive ty shifts it down. Pixels shifted outside the image boundary are filled with black.",
   },
 ];
--- a/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
@@ -96,15 +96,17 @@ export const geometricBlocks = [
   },
   {
     type: "geometric_affineimage",
-    message0: "Apply affine transformation | translate x %1 y %2",
-    args0: [
-      { type: "field_number", name: "tx", value: 50 },
-      { type: "field_number", name: "ty", value: 100 },
+    message0: "Apply affine transformation %1",
+    args0: [{ type: "input_dummy" }],
+    message1: "translate x %1 y %2",
+    args1: [
+      { type: "field_number", name: "tx", value: 50, min: -10000, max: 10000, precision: 1 },
+      { type: "field_number", name: "ty", value: 100, min: -10000, max: 10000, precision: 1 },
     ],
     previousStatement: null,
     nextStatement: null,
     style: "geometric_style",
     tooltip:
-      "Applies an affine transformation to the image - Translates the image by the specified number of pixels along the X and Y axes. Positive tx shifts the image right; positive ty shifts it down. Pixels shifted outside the image boundary are filled with black.",
+      "Applies a 2×3 affine transformation matrix to warp the image. Currently exposes translation (tx, ty); the underlying transform supports rotation, scaling, and shearing. Positive tx shifts the image right; positive ty shifts it down. Pixels shifted outside the image boundary are filled with black.",
   },
 ];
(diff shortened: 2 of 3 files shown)


imagelab-frontend/src/blocks/definitions/geometric.blocks.ts
```diff
@@ -96,11 +96,15 @@ export const geometricBlocks = [
   },
   {
     type: "geometric_affineimage",
-    message0: "Apply affine transformation",
+    message0: "Apply affine transformation | translate x %1 y %2",
+    args0: [
+      { type: "field_number", name: "tx", value: 50 },
+      { type: "field_number", name: "ty", value: 100 },
+    ],
```
ivantha: **[MEDIUM]** field_number inputs lack min, max, and precision constraints

The two `field_number` fields have no `min`, `max`, or `precision` properties. Blockly's `field_number` supports all three. Without them, a user can enter `999999999` (shifting the image completely off canvas producing an all-black result) or a very high-precision float that is serialised as a long string. Adding constraints closes the loop with the backend validation concern and improves discoverability of valid ranges.

**Suggestion:**
```
```typescript
{ type: "field_number", name: "tx", value: 50, min: -10000, max: 10000, precision: 1 },
{ type: "field_number", name: "ty", value: 100, min: -10000, max: 10000, precision: 1 },
```
The specific `min`/`max` bounds should reflect the maximum sensible translation relative to expected image sizes. If the project has a canvas size constant, use that.
```

File: imagelab-frontend/src/data/operatorDocs.ts [hunk 2]
```diff
@@ -69,14 +69,21 @@ export const operatorDocs: Record<string, OperatorDoc> = {
   },
   geometric_affineimage: {
     name: "Affine Transform",
```
ivantha: **[MEDIUM]** Operator description narrowed — name vs. capability mismatch

The operator is named `AffineImage` / "Affine Transform", but the updated description now reads *"Applies a 2x3 affine transformation matrix to **translate** the image."* The class name and the underlying `cv2.warpAffine` call are capable of any affine operation (rotation, shear, scale), yet the documentation now explicitly restricts the mental model to translation only. This creates a misleading API surface: a future developer extending the block to support rotation would find the docs actively contradicting what the code can do. The old description was actually more accurate.

**Suggestion:**
```
Keep the description general while still describing the current UI surface:

```ts
description: "Applies a 2×3 affine transformation matrix to warp the image. Currently exposes translation (tx, ty); the underlying transform supports rotation, scaling, and shearing.",
```
This documents what the block *does today* without misleading readers about the underlying capability.
```

File: imagelab-backend/app/operators/geometric/affine_image.py [hunk 3]
```diff
@@ -7,5 +7,7 @@
 class AffineImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         rows, cols = image.shape[:2]
-        M = np.float64([[1, 0, 50], [0, 1, 100]])
+        tx = float(self.params.get("tx", 50))
+        ty = float(self.params.get("ty", 100))
+        M = np.float64([[1, 0, tx], [0, 1, ty]])
         return cv2.warpAffine(image, M, (cols, rows))
```
ivantha: **[LOW]** No unit tests added for parameterised behavior

The PR description explicitly leaves "New tests added" unchecked. The most impactful logic change — reading dynamic `tx`/`ty` from `params` instead of hardcoded values — has no automated coverage. Three scenarios should minimally be tested: (1) default values (tx=50, ty=100) produce the same result as the old hardcode; (2) custom tx/ty values produce the expected shift; (3) invalid params (non-numeric, None) raise or fallback gracefully.

**Suggestion:**
```
```python
def test_affine_image_default_params():
    op = AffineImage(params={})
    result = op.compute(dummy_image)
    # assert pixel at (100, 50) equals source (0, 0)

def test_affine_image_custom_params():
    op = AffineImage(params={"tx": 10, "ty": 20})
    result = op.compute(dummy_image)
    # assert pixel at (20, 10) equals source (0, 0)

def test_affine_image_invalid_params():
    op = AffineImage(params={"tx": "bad", "ty": None})
    # assert falls back to defaults or raises ValueError cleanly
```
```

File: imagelab-frontend/src/blocks/definitions/geometric.blocks.ts [hunk 4]
```diff
@@ -96,11 +96,15 @@ export const geometricBlocks = [
   },
   {
     type: "geometric_affineimage",
-    message0: "Apply affine transformation",
+    message0: "Apply affine transformation | translate x %1 y %2",
```
ivantha: **[LOW]** Block message string uses pipe separator — verify renderer compatibility

`message0: "Apply affine transformation | translate x %1 y %2"` uses a `|` character. In Blockly, `|` is not a standard token in `message0` — it is treated as a literal character, not a row/section separator (that would require a `%` dummy input). If the intended goal was to visually split the label into two lines or sections, this won't work as expected and the block will display the pipe literally in the label text. If it was intentional as a visual delimiter, this should be documented.

**Suggestion:**
```
If the goal is a two-row block, use a `field_label` dummy input:
```typescript
message0: "Apply affine transformation %1",
args0: [{ type: "input_dummy" }],
message1: "translate x %1 y %2",
args1: [
  { type: "field_number", name: "tx", value: 50 },
  { type: "field_number", name: "ty", value: 100 },
],
```
If the pipe is simply a stylistic choice intended to render literally, add a comment explaining the decision.
```

ivantha (commented):
## ◎ Argus Review

This PR correctly exposes `tx`/`ty` translation parameters on `AffineImage`, replacing a hardcoded matrix, and preserves backward compatibility via matching defaults. The change is straightforward and the frontend/backend/docs are updated consistently. The primary concerns are a lack of defensive error handling around the `float()` conversion (which will raise an unhandled `ValueError` on invalid input), the absence of input bounds validation (which silently produces all-black images for extreme values), and magic constants duplicated across three locations. No new unit tests were added despite a behavioral change to the core computation path. The Blockly `field_number` fields also lack `min`/`max` constraints, leaving the door open for absurd user inputs.

### ⚠️ Breaking Changes

- **[INFO]** Saved workspace blocks missing tx/ty will silently use backend defaults, not frontend defaults: **Previous behavior**: The block had no serializable fields; `tx=50, ty=100` were hardcoded in Python.

**New behavior**: The block now serialises `tx` and `ty` as Blockly field values. Existing saved workspaces (XML/JSON) that contain `geometric_affineimage` blocks will not have `tx`/`ty` fields in their serialised form.

**Who is affected**: Users with previously saved pipelines that include the `AffineImage` block.

**Migration path**: When Blockly loads an old block definition without the new fields, it falls back to the `value` defaults in the block definition (50 and 100), which matches the old hardcoded values. Backend also defaults to 50/100 if params are absent. Functionally the behaviour is preserved. However, if the workspace serialization format stores field values at save time (not load time), re-saving will now persist `tx=50, ty=100` explicitly — this is a benign one-time migration. No manual migration is required, but this should be noted in release notes.
  > **Suggestion:** Add a brief note in the changelog or migration guide: *"Existing pipelines using `AffineImage` will continue to work unchanged. The block will now display `tx=50, ty=100` as editable fields after the first re-save."*


### PR Description Issues

- **[LOW]** PR description does not mention absence of new tests or rationale: The description acknowledges manual testing and that existing tests pass, but the `New tests added` checkbox is explicitly unchecked with no explanation. For a behavioral change that replaces hardcoded values with dynamic params, reviewers and future maintainers benefit from understanding *why* no new tests were added (e.g., "covered by integration tests", "tracked in issue #XYZ", "deferred to follow-up PR").
  > **Suggestion:** Add a brief note under the testing section, for example:

> Unit tests for the parameterised `tx`/`ty` logic are tracked in #XXX and will be added in a follow-up. The existing test suite exercises the operator at the integration level and continues to pass.
- **[INFO]** PR description quality is strong — minor enhancement opportunity: The description clearly states motivation (fixes #191), what changed (hardcoded matrix → params), default value preservation, and how it was tested. The type-of-change section is correctly marked. The only omission is the saved-workspace backward-compatibility note and the absent test rationale (covered in a separate finding). Overall this is a well-written PR description.
  > **Suggestion:** Consider adding a one-line backward compatibility note:

> **Backward compatibility**: Default values `tx=50, ty=100` are identical to the previous hardcoded values. Existing saved pipelines will continue to behave identically.

ivantha:
Reviewed. Replaces the hardcoded AffineImage translation matrix with user-configurable `tx`/`ty` fields, preserving the previous defaults (50, 100). The backend defensively handles invalid values (non-numeric, NaN, Inf) and clamps translation to image bounds. Frontend block definition and `operatorDocs` are updated to match. Merging via rebase.

ivantha:

xxivani:
Hi @ivantha, I've pushed the updated branch!

ivantha:
Thanks @xxivani! Closing this in favour of #82 which solves the same AffineImage hardcoded-params problem by exposing all 12 affine matrix coordinates — a more complete fix. Your tx/ty approach is solid; feel free to review #82 and contribute!

PR #215, opened 2026-03-14
fix: use convertScaleAbs in Laplacian to prevent uint8 wrapping

### LINKED ISSUE #214: [Bug] : Laplacian operator uses np.uint8() causing overflow wrapping instead of clipping
**Describe the bug**
The Laplacian operator converts its float output using `np.uint8(np.absolute(laplacian))`, which causes modulo wrapping instead of saturating conversion.

When Laplacian responses exceed 255 (which commonly happens at edges), `np.uint8()` wraps values modulo 256 rather than clipping them. For example, a response of `600` becomes `88` instead of `255`. This can produce incorrect edge intensities and make strong edges appear weaker than smaller ones.

File:
imagelab-backend/app/operators/transformation/laplacian.py

Relevant code:

return np.uint8(np.absolute(laplacian))

The Sobel and Scharr operators in the same codebase already use `cv2.convertScaleAbs()`, which correctly applies absolute value and saturating conversion.

**To Reproduce**

2. Upload any image with clear edges (text, buildings, objects, etc.).
3. Add an **Apply Laplacian** block to the pipeline.
4. Execute the pipeline.

Result:
Strong edges may appear dim or inconsistent because values greater than 255 wrap around instead of clipping.

**Expected behavior**
The Laplacian output should convert the float response to `uint8` using saturating clipping so that strong edges appear bright and edge strength ordering is preserved.

Using `cv2.convertScaleAbs()` would produce the correct behavior and align the implementation with the Sobel and Scharr operators.

**Screenshots**
If needed, visual examples of the artifact can be provided (e.g., strong edges appearing darker than weaker edges).

**Environment:**
- OS: Ubuntu 22.04
- Browser: Chrome
- Node.js version: 18.x
- Python version: 3.12

**Additional context**
The issue appears to be limited to the Laplacian operator. The Sobel and Scharr implementations already use `cv2.convertScaleAbs(...)`, which correctly performs absolute value and saturation to `[0,255]`. Updating the Laplacian operator to use the same conversion would resolve the issue and keep operator behavior consistent across the codebase.

anusha19murthy:
## Description

Fixes incorrect output conversion in the Laplacian operator.

The implementation in `imagelab-backend/app/operators/transformation/laplacian.py` was converting float results using `np.uint8(np.absolute(laplacian))`. This causes modulo wrapping when values exceed 255 instead of clipping them, which can produce incorrect edge intensities (e.g., a response of 600 becomes 88 instead of 255).

This PR replaces that conversion with `cv2.convertScaleAbs(laplacian)`, which performs absolute value and saturating conversion to `[0,255]`. This ensures correct edge strength representation and keeps the behavior consistent with the Sobel and Scharr operators already present in the codebase.

Fixes #214

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Verified that the Laplacian operator runs correctly on sample images.
- Confirmed that strong edges no longer wrap around due to uint8 overflow.
- Ran the existing test suite locally to ensure no regressions.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

N/A

## Checklist

- [x] Tests pass locally

Note: this fix is fully effective because the existing implementation uses ddepth=cv2.CV_64F, which preserves signed Laplacian responses before the saturating conversion via cv2.convertScaleAbs.

### DIFF
--- a/imagelab-backend/app/operators/transformation/laplacian.py
+++ b/imagelab-backend/app/operators/transformation/laplacian.py
@@ -20,5 +20,11 @@ class Laplacian(BaseOperator):
         if ksize not in _VALID_KSIZE:
             raise ValueError(f"Invalid ksize={ksize}; must be one of {sorted(_VALID_KSIZE)}")
 
+        # ddepth=cv2.CV_16S preserves signed Laplacian responses (both positive and negative)
+        # before the saturating cast. Using CV_8U would clip values during computation itself,
+        # masking large overshoots and making convertScaleAbs ineffective.
         laplacian = cv2.Laplacian(image, ddepth, ksize=ksize)
-        return np.uint8(np.absolute(laplacian))
+        # cv2.convertScaleAbs applies abs() and clips to [0, 255] (saturating cast).
+        # Using np.uint8() directly would wrap values > 255 modulo 256, producing
+        # incorrect edge intensities (e.g., a response of 600 would become 88).
+        return cv2.convertScaleAbs(laplacian)
--- a/imagelab-backend/tests/operators/transformation/test_laplacian.py
+++ b/imagelab-backend/tests/operators/transformation/test_laplacian.py
@@ -1,0 +1,14 @@
+import numpy as np
+
+from app.operators.transformation.laplacian import Laplacian
+
+
+def test_laplacian_no_uint8_wrapping():
+    """Regression test: strong edges must be clipped to 255, not wrapped modulo 256."""
+    img = np.zeros((20, 20), dtype=np.uint8)
+    img[10:, :] = 255  # sharp horizontal edge
+    operator = Laplacian({})
+    result = operator.compute(img)
+    assert result.dtype == np.uint8, "Output must be uint8"
+    assert result.max() == 255, "Expected saturated edge response of 255"
+    assert result[5, 0] == 255 or result[10, 0] == 255, "Edge pixel should be saturated to 255"


ivantha (commented):
## ◎ Argus Review

This is a small, targeted, and **correct** bug fix. The replacement of `np.uint8(np.absolute(laplacian))` with `cv2.convertScaleAbs(laplacian)` eliminates uint8 modulo-wrapping for strong edge responses and aligns the Laplacian operator with the established pattern used by the Sobel and Scharr operators in the same codebase. The PR description is clear and well-motivated. The primary gap is the absence of an automated regression test specifically asserting saturating (clipped) behavior — without such a test, this bug could silently re-appear in a future refactor. A secondary minor point is verifying that `ddepth` is a signed type so the full Laplacian response range is available for `convertScaleAbs` to work correctly.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: output values for strong edges now clip instead of wrap: **Previous behavior**: Pixels with absolute Laplacian response > 255 wrapped modulo 256. For example, a response of 600 produced an output pixel value of 88; a response of 256 produced 0.

**New behavior**: Those same pixels now saturate to 255. A response of 600 produces 255; a response of 256 also produces 255.

**Who is affected**: Any downstream code, model, or pipeline that was — intentionally or accidentally — calibrated to the old (incorrect) wrapped output values. In practice this is unlikely to be intentional, so this is a bug fix rather than a breaking change in the conventional sense.

**Migration path**: No migration is required for correct usage. If any downstream test assertions were written against the old wrapped values, those assertions should be updated to reflect the now-correct clipped values.
  > **Suggestion:** No migration action needed. This is a correctness fix. Update any test fixtures that snapshot raw Laplacian output values to reflect the new clipped range.


### PR Description Issues

- **[INFO]** PR description is thorough and well-written: The PR description does an excellent job of:
- Clearly identifying the root cause (modulo wrapping in `np.uint8`).
- Providing a concrete numeric example (`600 → 88` vs `600 → 255`).
- Justifying the chosen fix by referencing consistency with Sobel/Scharr operators.
- Being transparent about testing: the "New tests added" checkbox is honestly left unchecked.

No significant gaps in the description itself.
  > **Suggestion:** The description is solid. One optional improvement: explicitly mention the `ddepth` value used in the operator, since `convertScaleAbs` only fully benefits when a signed depth is used. Example addition:

> Note: this fix is fully effective because the existing implementation uses `ddepth=cv2.CV_16S` (or equivalent), which preserves signed Laplacian responses before the saturating conversion.

PR #217, opened 2026-03-15
fix(filtering): correct swapped width/height kernel size in BoxFilter

### LINKED ISSUE #216: [Bug] BoxFilter passes swapped (height, width) kernel size to cv2.boxFilter()
### Describe the bug
In `imagelab-backend/app/operators/filtering/box_filter.py`, the kernel size is passed as `(height, width)` to `cv2.boxFilter()`, but OpenCV expects `(width, height)`.

### Relevant code
```python
return cv2.boxFilter(
    image,
    depth,
    (height, width),  # ← swapped!
    ...
)
```

### Expected behavior
Should be `(width, height)` to match OpenCV's ksize convention:
```python
return cv2.boxFilter(
    image,
    depth,
    (width, height),  # ← correct
    ...
)
```

### Additional context
The same swap bug was already fixed in the Blur operator (issue #122). This fix aligns BoxFilter with the same convention used across the codebase.

### Environment
- Python 3.x
- OpenCV (cv2)

ajithhraj:
## Description

Fixed `BoxFilter` silently applying a transposed kernel for non-square kernel sizes.

`cv2.boxFilter()` expects `ksize` as `(width, height)`, but the implementation
passed `(height, width)`. For square kernels (`width == height`) this had no
visible effect; for asymmetric kernels the wrong axis was smoothed.

This issue was identified while checking operator consistency after the earlier
Blur operator fix.

Fixes #216

## Impact

- Non-square kernels (`width != height`): output now changes to the correct result
- Square kernels (`width == height`): no behavior change

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- Manual verification against OpenCV's documented `ksize` convention
- Added an automated regression test for the asymmetric-kernel case
- Ran:

```bash
python -m pytest imagelab-backend/tests/test_filtering_operators.py
python -m py_compile imagelab-backend/app/operators/filtering/box_filter.py imagelab-backend/tests/test_filtering_operators.py

### DIFF
--- a/imagelab-backend/app/operators/filtering/box_filter.py
+++ b/imagelab-backend/app/operators/filtering/box_filter.py
@@ -1,6 +1,5 @@
 import cv2
 import numpy as np
-
 from app.operators.base import BaseOperator
 
 
@@ -14,7 +13,7 @@ class BoxFilter(BaseOperator):
         return cv2.boxFilter(
             image,
             depth,
-            (height, width),
+            (width, height),  # OpenCV ksize convention: (width, height)
             anchor=(point_x, point_y),
             normalize=True,
             borderType=cv2.BORDER_DEFAULT,
--- a/imagelab-backend/app/operators/filtering/box_filter.py
+++ b/imagelab-backend/app/operators/filtering/box_filter.py
@@ -1,5 +1,6 @@
 import cv2
 import numpy as np
+
 from app.operators.base import BaseOperator
 
 
--- a/imagelab-backend/tests/test_filtering_operators.py
+++ b/imagelab-backend/tests/test_filtering_operators.py
@@ -1,3 +1,4 @@
+import cv2
 import numpy as np
 import pytest
 
@@ -74,6 +75,24 @@ class TestBoxFilter:
     def test_output_is_uint8(self, color_image):
         result = BoxFilter({"depth": -1}).compute(color_image)
         assert result.dtype == np.uint8
+
+    def test_asymmetric_kernel_uses_width_height_convention(self):
+        image = np.arange(75, dtype=np.uint8).reshape(5, 5, 3)
+        width, height = 1, 5
+
+        result = BoxFilter({"width": width, "height": height, "depth": -1}).compute(
+            image
+        )
+        expected = cv2.boxFilter(
+            image,
+            -1,
+            (width, height),
+            anchor=(-1, -1),
+            normalize=True,
+            borderType=cv2.BORDER_DEFAULT,
+        )
+
+        np.testing.assert_array_equal(result, expected)
 
 
 # Sharpen

### CODE REVIEW

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR correctly fixes a real bug: `cv2.boxFilter()` expects `ksize` as `(width, height)`, and the previous code had the arguments transposed. The core one-line change is accurate, well-commented, and consistent with the earlier `Blur` operator fix in #122. However, no automated regression test accompanies the fix, which is a significant gap for a silent, visually-plausible bug that could easily be reintroduced. Two minor issues round out the review: an incidental PEP 8 violation (removal of the blank line between import groups) that is unrelated to the bug fix, and a PR description that could better communicate the impact scope (only non-square kernels are affected) and the testing gap.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: non-square BoxFilter kernels will produce different output: **Previous behavior:** `cv2.boxFilter()` received `(height, width)` as `ksize`, so a `BoxFilter(width=3, height=7)` configuration actually applied a *7-wide × 3-tall* kernel.

**New behavior:** `cv2.boxFilter()` now correctly receives `(width, height)`, so the same configuration applies a *3-wide × 7-tall* kernel — the intended shape.

**Who is affected:** Any end-users or downstream pipelines that have tuned `BoxFilter` parameters for non-square kernels (`width ≠ height`) based on the previous (incorrect) behavior. Their outputs will change after this fix.

**Square-kernel users are unaffected** (`width == height` makes the transposition a no-op).

**Migration path:** Users relying on the (buggy) swapped behavior should swap their `width` and `height` parameter values in their pipeline configurations to preserve the old kernel shape. For example, a previous configuration of `width=3, height=7` that was effectively applying a 7×3 kernel should be updated to `width=7, height=3` if the 7×3 shape was intentional.
  > **Suggestion:** Add a note to the changelog or release notes flagging this as a corrected behavior with a migration hint for users who configured asymmetric kernels.


### PR Description Issues

- **[MEDIUM]** PR description lacks motivation context and impact scope: The description correctly identifies the fix and references the OpenCV convention, but it is missing:

1. **Impact scope**: No mention that only *non-square* kernels are affected (square kernels are unchanged). This would help reviewers and users gauge the blast radius.
2. **Root cause context**: No explanation of how the bug was discovered (was it observed visually? reported by a user? noticed during code review of #122?).
3. **Testing depth**: "Manual testing" is vague. It does not clarify *which* kernel shapes were tested, whether a side-by-side output comparison was done, or whether automated tests were considered and intentionally omitted.
  > **Suggestion:** Suggested improved description:
```markdown
## Description
Fixed `BoxFilter` silently applying a transposed kernel for non-square kernel sizes.
`cv2.boxFilter()` expects `ksize` as `(width, height)`, but the implementation
passed `(height, width)`. For square kernels (`width == height`) this had no
visible effect; for asymmetric kernels the wrong axis was smoothed.

Discovered during review of operator consistency following the `Blur` fix in #122.

Fixes #216

## Impact
- **Non-square kernels** (`width ≠ height`): output will change to the correct result.
- **Square kernels** (`width == height`): no change in behavior.

## How Has This Been Tested?
- Manually verified with a test image using `width=3, height=7` and confirmed the
  output now matches the expected horizontal-blur appearance.
- A regression test covering the asymmetric-kernel case should be added in a
  follow-up (tracked in #XYZ).
```
- **[LOW]** PR description omits follow-up action for automated testing gap: The "How Has This Been Tested?" section only checks the "Manual testing" box and provides a brief rationale about convention alignment. It does not acknowledge the absence of an automated test, nor does it commit to adding one. This leaves the bug regression protection entirely to human vigilance.
  > **Suggestion:** Either add a follow-up ticket reference acknowledging the missing test, or include the test in this PR. At minimum, the description should state:
> *Note: an automated regression test for asymmetric kernel correctness will be added in issue #XYZ to prevent this class of bug from recurring.*

PR #237, opened 2026-03-16
Fix uint8 overflow wrapping in Laplacian operator output

### LINKED ISSUE #236: [fix] Laplacian operator incorrectly converts gradients using np.uint8, causing overflow wrapping
**Describe the bug**
The Laplacian operator converts its output using:

```
np.uint8(np.absolute(laplacian))
```

in `imagelab-backend/app/operators/transformation/laplacian.py`.

When gradient magnitudes exceed 255 (which commonly happens at strong edges), `np.uint8()` performs modulo-256 wrapping instead of saturating the value.

Example:

| Actual gradient | Output with `np.uint8()` | Correct output |
|----------------|-------------------------|---------------|
| 256 | 0 | 255 |
| 300 | 44 | 255 |
| 400 | 144 | 255 |

This silently corrupts edge intensities. Strong edges may appear darker than weaker edges, producing incorrect Laplacian edge detection results.

This behavior is also inconsistent with the Sobel and Scharr operators in the same codebase, which correctly use `cv2.convertScaleAbs()` to perform saturating conversion.

---

**To Reproduce**

2. Upload any image with strong edges (text, shapes, buildings, etc.).
3. Add a **Laplacian** operator to the pipeline.
4. Execute the pipeline.
___
**Expected behavior**
Gradient magnitudes should be clipped to the `[0,255]` range instead of wrapping around. Strong edges should appear brighter than weaker ones.
___
**Screenshots**
Optional: artifacts appear as dark or banded edges where strong gradients should be bright.
___
**Environment:**
- OS: Ubuntu 22.04
- Browser: Chrome
- Node.js version: 18.x
- Python version: 3.12
___
**Additional context**
The issue occurs because `np.uint8()` wraps values modulo 256.  
A safer conversion would be:

```
np.clip(np.absolute(laplacian), 0, 255).astype(np.uint8)
```

or using OpenCV's recommended approach:

```
cv2.convertScaleAbs(laplacian)
```

which is already used by the Sobel and Scharr operators in the ImageLab codebase.

anusha19murthy:
## Description

Fix incorrect conversion of Laplacian output values.

The Laplacian operator previously converted gradients using:

```
np.uint8(np.absolute(laplacian))
```

This causes **modulo-256 wrapping** when values exceed 255. As a result, strong edge gradients could wrap to incorrect low values, producing inconsistent or misleading edge intensities.

This PR replaces the conversion with a saturating conversion:

```
np.clip(np.absolute(laplacian), 0, 255).astype(np.uint8)
```

This ensures gradient magnitudes are properly clipped to the `[0,255]` range instead of wrapping.

Fixes #236
___
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
___
## How Has This Been Tested?

- Ran the existing test suite locally.
- Verified Laplacian output on sample images with strong edges.
- Confirmed gradients above 255 are clipped instead of wrapping.

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing
___
N/A
___
## Checklist

- [x] Tests pass locally

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,18 +1,16 @@
 name: Lint
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [main]
   push:
     branches: [main]
-  issue_comment:
-    types: [created]
 
 concurrency:
   group: >-
-    lint-${{
-      github.event_name == 'issue_comment'
-        && format('pr-{0}', github.event.issue.number)
+    ci-${{
+      github.event_name == 'pull_request_target'
+        && format('pr-{0}', github.event.pull_request.number)
         || github.ref
     }}
   cancel-in-progress: true
@@ -24,21 +22,18 @@ permissions:
 jobs:
   frontend-lint:
     name: Frontend Lint
-    if: >-
-      github.event_name != 'issue_comment'
-      || (
-        github.event.issue.pull_request
-        && contains(github.event.comment.body, '/lint')
-        && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
-      )
     runs-on: ubuntu-latest
+    outputs:
+      eslint: ${{ steps.eslint.outcome }}
+      prettier: ${{ steps.prettier.outcome }}
     defaults:
       run:
         working-directory: imagelab-frontend
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ github.event_name == 'issue_comment' && format('refs/pull/{0}/head', github.event.issue.number) || '' }}
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ github.event.pull_request.head.sha || '' }}
       - uses: actions/setup-node@v4
         with:
           node-version: 20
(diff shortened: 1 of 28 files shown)


ivantha (commented):
## ◎ Argus Review

This is a small, focused, and correct bug fix. The one-line change from `np.uint8(np.absolute(laplacian))` to `np.clip(np.absolute(laplacian), 0, 255).astype(np.uint8)` properly addresses the well-known NumPy uint8 modulo-wrapping issue for Laplacian gradients exceeding 255. The fix itself is technically sound, idiomatic, and addresses a real correctness problem that would silently corrupt strong-edge pixel values. The primary concerns are the absence of a targeted regression test (which should be added to prevent future regressions) and a minor redundancy in clipping at 0 after `np.absolute`. The PR description is clear and accurate, though it would benefit from a rationale for skipping new tests and a docstring update to document the saturating conversion behavior.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: strong-edge pixel values will differ from previous output: **Previous behavior**: Any Laplacian absolute gradient value `v > 255` was converted as `v % 256`, causing values like 256→0, 300→44, 510→254. Consumers receiving this output would see low pixel intensities for very strong edges.

**New behavior**: Values `> 255` are clamped to `255`, and values `≤ 255` are unchanged. Strong-edge pixels that previously wrapped to low values will now correctly appear as bright (255).

**Who is affected**: Any downstream code, pipeline stage, model, or user workflow that ingested the Laplacian output and was (perhaps unknowingly) calibrated to the wrapped values. In practice, the previous behavior was a bug, so genuine intentional dependence on wrapping is very unlikely — but it is a real change in pixel-level output.

**Migration path**: No code changes should be required for consumers. If any downstream thresholding, normalization, or machine-learning inference was tuned on the buggy output, those parameters may need recalibration.
  > **Suggestion:** Document this behavioral change explicitly in the changelog or release notes so operators/users know that stored edge maps from previous versions will differ from newly computed ones.


### PR Description Issues

- **[LOW]** No new tests added — PR description should explain rationale: The `- [ ] New tests added` checkbox is left unchecked with no explanation. For a bug fix PR, this is a notable omission — reviewers and future maintainers are left without a regression guard. A brief justification (e.g., "existing test suite covers this path" or "test to be added in follow-up #NNN") would improve transparency.
  > **Suggestion:** Update the testing section to either:
1. Add a new test (preferred — see the testing finding above), then check the box.
2. Provide an explicit rationale for why no test was added:
```markdown
- [ ] New tests added — _Deferred to follow-up issue #NNN; existing integration tests exercise the Laplacian path but do not assert clipping behavior specifically._
```
- **[INFO]** Documentation update checkbox unchecked with no explanation: The `- [ ] I have added/updated documentation as needed` checkbox is unchecked. For a behavioral change (even a bug fix), it is worth confirming whether the operator has any docstring, user-facing documentation, or API reference that should note the corrected output behavior — especially since previous outputs from this operator will differ.
  > **Suggestion:** Either update the docstring in `laplacian.py` to describe the saturating uint8 conversion, or explicitly note in the PR that no documentation update is needed:
```markdown
- [x] I have added/updated documentation as needed — _No user-facing docs reference the conversion detail; updated inline docstring in `laplacian.py`._
```
Suggested docstring addition:
```python
    """
    Apply the Laplacian operator to detect edges.

    Gradient magnitudes are clipped to [0, 255] before conversion to uint8
    (saturating conversion) to avoid modulo-256 wrapping artefacts on
    strong edges.
    """
```

Issue #220, opened 2026-03-15
[Bug] ApplyBorders crashes on negative border values with no validation


ajithhraj:
### Describe the bug
The `ApplyBorders` operator in `imagelab-backend/app/operators/thresholding/apply_borders.py` does not validate that border values are non-negative before passing them to `cv2.copyMakeBorder()`.

Passing a negative value (e.g. `borderTop=-5`) causes OpenCV to throw an error and crash the pipeline.

### Relevant code
```python
top = int(self.params.get("borderTop", 0))
bottom = int(self.params.get("borderBottom", 0))
left = int(self.params.get("borderLeft", 0))
right = int(self.params.get("borderRight", 0))
return cv2.copyMakeBorder(image, top, bottom, left, right, cv2.BORDER_CONSTANT, value=[0, 0, 0])
```

### Expected behavior
Negative border values should be clamped to 0 before passing to `cv2.copyMakeBorder()`:
```python
top = max(0, int(self.params.get("borderTop", 0)))
bottom = max(0, int(self.params.get("borderBottom", 0)))
left = max(0, int(self.params.get("borderLeft", 0)))
right = max(0, int(self.params.get("borderRight", 0)))
```

### Additional context
Issue #193 already tracks the hardcoded border type. This is a separate bug about missing input validation.

ivantha:
Closing: this is already fixed on `main`.

ApplyBorders now clamps every border side with max(0, int(...)) and raises a clear ValueError on non-numeric input before calling cv2.copyMakeBorder.

Evidence: imagelab-backend/app/operators/thresholding/apply_borders.py:18-29 (commit 1519f55).

Thanks for the report. If you still see this on `main`, comment here and we will reopen.

Issue #120, opened 2026-03-01
[Bug] ScaleImage allows fx=0/fy=0 and crashes pipeline


Jayant-kernel:
**Describe the bug**
`ScaleImage` accepts `fx=0` or `fy=0` (UI minimum is `0`) and then crashes during execution in OpenCV `resize`.

**To Reproduce**

1. Upload any image in ImageLab.
2. Add `geometric_scaleimage` block.
3. Set `fx = 0` (or `fy = 0`).
4. Run the pipeline.
5. See pipeline failure from `cv2.resize` assertion.

API-level repro (same behavior):

```python
from app.operators.geometric.scale_image import ScaleImage
ScaleImage({"fx": 0, "fy": 1}).compute(image)
```

**Expected behavior**
The operator should not crash.

Expected handling options:
- validate `fx` and `fy` as `> 0` and return a clear validation error, or
- clamp invalid values to safe defaults.

**Environment:**
- OS: Windows 11 (local repro)
- Browser: N/A (also reproducible directly in backend operator)
- Node.js version: N/A
- Python version: 3.14 (local repro), OpenCV 4.13.0

**Additional context**
Likely relevant code paths:
- `imagelab-frontend/src/blocks/definitions/geometric.blocks.ts` (`fx`/`fy` min currently `0`)
- `imagelab-backend/app/operators/geometric/scale_image.py` (passes zero scale to `cv2.resize`)

OpenCV error observed:
`(-215:Assertion failed) inv_scale_x > 0 in function 'cv::resize'`

PR #247, opened 2026-03-18
fix(drawing): clamp negative radius to 0 in DrawCircle

Kasturijadhav14:
The `DrawCircle` operator did not validate the `radius` parameter before
passing it to `cv2.circle()`. Passing a negative value (e.g. `radius=-5`)
caused OpenCV to throw an error and crash the entire pipeline with no
useful error message.

Clamped `radius` to a minimum of `0` using `max(0, ...)` — so instead
of crashing, the operator silently draws a zero-radius circle (a point):

python
radius = int(self.params.get("radius", 5))

python
radius = max(0, int(self.params.get("radius", 5)))

This follows the same defensive validation pattern already applied in
`ApplyBorders` (#220) and `ScaleImage` (#120). `DrawCircle` should
be consistent with those operators. 🔵

### DIFF
--- a/imagelab-backend/app/operators/drawing/draw_circle.py
+++ b/imagelab-backend/app/operators/drawing/draw_circle.py
@@ -9,7 +9,7 @@ class DrawCircle(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         result = image.copy()
         thickness = int(self.params.get("thickness", 2))
-        radius = int(self.params.get("radius", 5))
+        radius = max(0, int(self.params.get("radius", 5)))
         color = hex_to_bgr(self.params.get("rgbcolors_input", "#2828cc"))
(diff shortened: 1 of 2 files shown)


ivantha:
Thanks for the PR! The **Backend Lint** check is currently failing. Could you fix the linting errors (run `uv run ruff check . && uv run ruff format --check .` from `imagelab-backend/`) and push the updated branch? Happy to merge once CI is fully green.

ivantha:
Thanks @Kasturijadhav14! Closing this in favour of #291 which fixes the same DrawCircle negative-radius issue and is rebased on latest main with all CI checks passing. Appreciate the contribution — feel free to follow up on #291!

Kasturijadhav14:
Thanks for the update! Happy to follow along.

PR #252, opened 2026-03-19
Fix uint8 overflow in Laplacian operator and add regression test

### LINKED ISSUE #214: [Bug] : Laplacian operator uses np.uint8() causing overflow wrapping instead of clipping

anusha19murthy:
## Description

Fixes incorrect uint8 conversion in the Laplacian operator that caused silent overflow and corrupted edge intensities.

Previously, the implementation used:
np.uint8(np.absolute(laplacian))

This wraps values modulo 256, causing strong edges (values >255) to appear darker or incorrect. The fix replaces this with:
cv2.convertScaleAbs(laplacian)

which performs proper scaling + saturation, ensuring correct and monotonic edge intensity representation.

Also adds a regression test to prevent this issue from reappearing.

Fixes #214 

---

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---

## How Has This Been Tested?

- Verified output dtype is uint8 and values are within [0, 255]
- Confirmed strong edges are brighter than weaker ones (no wrap-around)
- Tested with multiple kernel sizes (ksize=1, 3)
- Ran full backend test suite locally

- [x] Existing tests pass
- [x] New tests added
- [x] Manual testing

---

N/A

---

## Checklist

- [x] I have added regression tests for the fix
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/conversions/brightness_and_contrast.py
+++ b/imagelab-backend/app/operators/conversions/brightness_and_contrast.py
@@ -1,0 +1,42 @@
+import numpy as np
+
+from app.operators.base import BaseOperator
+
+
+class BrightnessAndContrast(BaseOperator):
+    """
+    Adjusts the brightness and contrast of an image.
+
+    Params:
+        brightnessValue (int): Additive brightness offset [-100, 100]. Clamped if out of range.
+        contrastValue (float): Multiplicative contrast scale [0.0, 3.0]. 1.0 = no change.
+    """
+
+    def compute(self, image: np.ndarray) -> np.ndarray:
+        """
+        Apply brightness and contrast adjustment.
+
+        Args:
+            image: Input uint8 NumPy ndarray (grayscale or multi-channel).
+
+        Returns:
+            Adjusted image as a uint8 NumPy ndarray of the same shape.
+        """
+        if image is None or not isinstance(image, np.ndarray) or image.size == 0:
+            raise ValueError("BrightnessAndContrast.compute: received an invalid or empty image")
+        try:
+            brightness = int(self.params.get("brightnessValue", 0))
+        except (TypeError, ValueError):
+            brightness = 0
+        try:
+            contrast = float(self.params.get("contrastValue", 1.0))
+        except (TypeError, ValueError):
+            contrast = 1.0
+
+        # Ensure brightness and contrast values are within valid ranges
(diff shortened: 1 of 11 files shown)


anusha19murthy:
@ivantha please review this pr once.

ivantha:
Reviewed. This is a correct and minimal fix — replacing `np.uint8(np.absolute(laplacian))` with `cv2.convertScaleAbs(laplacian)` properly handles overflow via saturation rather than modulo wrapping, and the regression tests validate both correct dtype/range and monotonic edge intensity. All CI checks pass. Merging via rebase.

ivantha:

PR #253, opened 2026-03-20
poc: add per-step intermediate previews with histogram analysis

anusha19murthy:
## Description

Traditional image processing tools like ImageLab return only the final output image after a pipeline executes. When a student chains 8 operators together and the result looks wrong, they have no way to identify which step caused the problem. Additionally, there is no quantitative feedback, no pixel statistics, no channel distributions making it impossible to understand the numerical impact of each operation. The existing pipeline executor is a black box: input goes in, final image comes out, everything in between is invisible.

---

## Solution

I introduced a compute_histogram() utility and extended the pipeline executor with an opt-in intermediate capture loop. When include_intermediates: true is passed in the request, the executor captures a base64 snapshot and per channel RGB histogram (mean, std, min, max + bin counts) after every operator.compute() call. I return the results as a structured intermediates array alongside the existing final output completely bypassing any changes to the operator implementations themselves. I preserved the existing API contract fully for all callers that do not set the flag.

---

## What This PoC Demonstrates

- Non-intrusive capture: I collect intermediate snapshots in the executor loop without modifying any of the 50+ existing operator classes.
- Opt-in design: include_intermediates defaults to false all existing API consumers are unaffected.
- Pixel-exact verification: I wrote integration tests that assert intermediate image values match sequential manual operator application.
- Partial results on failure: if an operator crashes mid-pipeline, I still return all intermediates captured before the failure.
- Zero new dependencies: I implemented everything entirely with opencv-python-headless and numpy, which are already in `pyproject.toml`.
- Strong foundation: this capture loop is the single shared infrastructure that all 5 GSoC features (previews, histograms, persistence, batch, macros) depend on.

---

## Testing

I wrote 18 tests that verify:

-` compute_histogram()` returns correct shape, channel count, and statistics for RGB and grayscale images.
- Histogram bin counts sum to total pixel count.
- Statistics (mean, std, min, max) match numpy ground truth.
- Pipeline with` include_intermediates: false` returns identical response to original API.
- Pipeline with` include_intermediates: true` returns one intermediate per operator step.
- Intermediate image pixel values are pixel-exact matches to manual sequential execution.
- Partial results are returned when an operator raises mid-pipeline.
- Response schema validates correctly via Pydantic.

All 18 new tests pass. All previously-passing tests still pass. Ruff lint + format clean.

---

## Running It

```
cd imagelab-backend
uv run pytest tests/test_histogram.py tests/test_intermediate_previews.py -v
```

---

## Next Steps

This PoC proves the core infrastructure works. In the full GSoC implementation I would:

- Add StepPreviewGallery and HistogramPanel React components to render intermediates in the frontend.
- Implement SSE streaming to avoid large single-response payloads for long pipelines.
- Add pipeline persistence (SQLModel Pipeline table + FastAPI CRUD routes) using the existing Alembic setup.
- Wrap the executor in a concurrent batch loop with progress tracking and ZIP download.
- Implement macro expansion so composite operators expand to their constituent steps before execution.

---

_This is a proof-of-concept for the GSoC 2026 "ImageLab: A Genuine Learning Environment for Image Processing" project_

### DIFF
--- a/imagelab-backend/POC.md
+++ b/imagelab-backend/POC.md
@@ -1,0 +1,60 @@
+# PoC: Per-Step Intermediate Previews with Histogram Analysis
+
+## What this demonstrates
+
+This proof of concept implements the **core mechanism** of the GSoC project: capturing the image state and computing per-channel histogram statistics **after every operator** in an ImageLab pipeline.
+
+Currently, ImageLab's pipeline executor processes a chain of OpenCV operators and returns only the **final** image. Students see no intermediate states, making the tool a black box. This PoC modifies the executor loop so that — when opt-in via `include_intermediates: true` — each completed step also produces:
+
+- A base64-encoded snapshot of the image at that point
+- Per-channel (B/G/R or Gray) histogram data (256 bins) with statistics (mean, std, min, max)
+
+This is the foundation that all other proposed features (histogram UI, pipeline persistence, batch processing, macros) build on.
+
+## What was changed
+
+**New file — `app/utils/histogram.py`**
+Computes per-channel histograms and pixel statistics for any `np.ndarray` image using `cv2.calcHist` and numpy. Handles both grayscale (1 channel) and colour (3 channels, BGR).
+
+**Modified — `app/models/pipeline.py`**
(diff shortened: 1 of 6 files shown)


ivantha:

PR #283, opened 2026-04-02
fix(tests): blurring and contour detection tests asserting wrong behaviour

### LINKED ISSUE #282: [BUG] Fix incorrect test expectations in blurring and contour detection tests
## **Description:**

Two test files have been testing the wrong things for a while and nobody caught it.

In `test_blurring_operators.py`, three tests are broken. The even-kernel tests for `GaussianBlur` and `MedianBlur` expect the operators to silently correct an even size to the next odd — that used to be the behaviour but both operators now just raise a `ValueError` instead. Same story with `test_kernel_size_1_is_identity` for `MedianBlur`, it passes `kernelSize=1` but the validator rejects anything below 3. All three need to be rewritten to assert the error gets raised.

In `test_contour_detection.py` the parametrized test is passing `color`, `retrieval_mode`, and `approximation_method` as param keys but the operator actually reads `rgbcolors_input`, `mode`, and `method`. The wrong keys just get silently ignored so the tests were never actually hitting those code paths — they were passing for the wrong reason. One case also uses `"LIST"` as a mode value which doesn't exist in `_MODE_MAP`. On top of that the shape assertions in `test_contour_detection_supports_gray_bgr_bgra` are wrong — grayscale input comes out as BGR `(H, W, 3)` not `(H, W)`, and BGRA comes out as BGRA `(H, W, 4)` not `(H, W, 3)`.

alikhere:
## Description

The blurring operators were updated at some point to validate kernel sizes strictly instead of silently correcting them, but the tests weren't updated to match. `test_even_kernel_corrected_to_odd` for both `GaussianBlur` and `MedianBlur` were trying to call the operators with even sizes and asserting the output equalled the next-odd result — but both operators now raise `ValueError` for even inputs. Same with `test_kernel_size_1_is_identity` for `MedianBlur`: the validator rejects `kernelSize=1` (min is 3). Rewrote all three to assert the `ValueError` is raised, which is what the validators actually guarantee.

The contour detection test had two separate problems. First, the parametrized cases were passing the wrong key names — `color`, `retrieval_mode`, `approximation_method` instead of the operator's actual param keys `rgbcolors_input`, `mode`, `method`. The wrong keys were silently ignored so the tests were passing while exercising nothing. Also removed the `LIST` mode case which isn't in `_MODE_MAP` at all. Second, the channel support test had wrong shape assertions: grayscale input gets converted to BGR for drawing (output is `(H, W, 3)`, not `(H, W)`), and BGRA input is preserved as-is (output `(H, W, 4)`, not `(H, W, 3)`).

Also includes the `color.py` ruff formatting fix (double blank line after import) that main still carries so CI doesn't fail on the unrelated pre-existing violation.

Fixes #282

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?
Ran `uv run pytest` — all previously broken tests now pass with correct assertions. Ran `uv run ruff check .` and `uv run ruff format --check .` with no errors.
- [x] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

N/A

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/tests/operators/filtering/test_contour_detection.py
+++ b/imagelab-backend/tests/operators/filtering/test_contour_detection.py
@@ -26,9 +26,8 @@ def make_rect_image(channels=3, bg=0, fg=255):
     "params",
     [
         {},  # defaults
-        {"color": "#00FF00", "thickness": 1, "retrieval_mode": "EXTERNAL", "approximation_method": "SIMPLE"},
-        {"color": "#FF0000", "thickness": 2, "retrieval_mode": "TREE", "approximation_method": "NONE"},
-        {"color": "#0000FF", "thickness": 3, "retrieval_mode": "LIST", "approximation_method": "SIMPLE"},
+        {"rgbcolors_input": "#00FF00", "thickness": 1, "mode": "EXTERNAL", "method": "SIMPLE"},
+        {"rgbcolors_input": "#FF0000", "thickness": 2, "mode": "TREE", "method": "NONE"},
     ],
 )
 def test_contour_detection_returns_valid_image(params):
@@ -50,8 +49,10 @@ def test_contour_detection_supports_gray_bgr_bgra(channels):
 
     result = op.compute(image.copy())
 
-    if channels == 4:
-        assert result.shape == (100, 100, 3)  # BGRA -> BGR conversion path
+    if channels == 1:
+        assert result.shape == (100, 100, 3)  # grayscale -> BGR conversion for drawing
+    elif channels == 4:
+        assert result.shape == (100, 100, 4)  # BGRA preserved, alpha retained
     else:
         assert result.shape == image.shape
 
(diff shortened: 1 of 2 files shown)


ivantha:
Reviewed. Fixes three categories of incorrect tests: blurring tests that expected silent correction of even/size-1 kernels but the validators now raise ValueError; contour detection tests that used wrong parameter key names (color/retrieval_mode/approximation_method instead of rgbcolors_input/mode/method) and incorrect output shape assertions for grayscale and BGRA inputs. All tests now accurately reflect the actual operator behavior. Merging via rebase.

ivantha:

alikhere:
@ivantha Thanks for the reviw, Rebased onto latest main,  PTAL

ivantha:

PR #286, opened 2026-04-06
 fix: ReadImage clear image only resetting the last registered block

### LINKED ISSUE #285: [BUG] `pipelineStore` stores `_imageResetFn` in Zustand state — anti-pattern, last-writer-wins bug
## **Description:**

`_imageResetFn` is stored as a plain Zustand state value. When multiple `ReadImage` blocks are created, each calls `registerImageReset(fn)`, and the last registration overwrites all previous ones. Only the most-recently-created block gets its label reset on `clearImage`. Functions in Zustand state also break React DevTools and serialisation. Replace with a `Set<() => void>` stored in a `useRef`-backed module-level ref or a proper subscriber pattern outside Zustand state.

**How to reproduce:**

1. Open ImageLab and add two ReadImage blocks to the workspace.
2. Load a different image into each block so both show a filename label.
3. Click the "Clear Image" button (or trigger `clearImage` from the store).
4. Observe that only one block's label resets to the placeholder — the other retains its filename.

**Expected behaviour:**

Both ReadImage blocks reset their labels when `clearImage` is called, regardless of the order they were created.

alikhere:
## Description

`_imageResetFn` was stored as a plain Zustand state value, so every `registerImageReset` call overwrote the previous one. With two `ReadImage` blocks, only the second one's label ever reset on "Clear Image". Functions in Zustand state also break React DevTools serialisation.

Replaced `_imageResetFn` with a module-level `Set<() => void>` that lives outside Zustand. `registerImageReset` now adds to the Set and returns an unsubscribe function. `clearImage` iterates the whole Set, so every live `ReadImage` block gets its label reset. `reset()` clears the Set so stale callbacks from a previous session don't survive a workspace reset.

The extension was updated to store the returned unsubscribe and call it inside the existing disposal Proxy, so the Set doesn't accumulate dead callbacks from deleted blocks.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Ran unit tests, integration tests, and manual testing in the browser with and without the backend running. Verified that with two `ReadImage` blocks on the workspace, clicking "Clear Image" now resets both blocks' labels instead of only the last registered one.

- [x] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Related issue:
Fixes #285 

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-frontend/src/blocks/extensions/readImageExtension.ts
+++ b/imagelab-frontend/src/blocks/extensions/readImageExtension.ts
@@ -39,8 +39,7 @@ function initReadImageBlock(block: Blockly.Block) {
     });
   }
 
-  // Register a reset callback when the image is cleared
-  usePipelineStore.getState().registerImageReset(() => {
+  const unregisterImageReset = usePipelineStore.getState().registerImageReset(() => {
     const label = block.getField("filename_label");
     if (label) label.setValue("No image");
   });
@@ -49,6 +48,7 @@ function initReadImageBlock(block: Blockly.Block) {
   block.dispose = new Proxy(block.dispose, {
     apply(target, thisArg, args) {
       fileInput.remove();
+      unregisterImageReset();
       return Reflect.apply(target, thisArg, args);
     },
   });
--- a/imagelab-frontend/src/store/pipelineStore.ts
+++ b/imagelab-frontend/src/store/pipelineStore.ts
@@ -3,6 +3,8 @@ import * as Blockly from "blockly";
 import { categories } from "../blocks/categories";
 import type { PipelineTimings } from "../types/pipeline";
 
+const imageResetListeners = new Set<() => void>();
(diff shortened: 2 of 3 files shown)


ivantha:
Reviewed. Fixes a bug where only the last registered ReadImage block had its label reset on 'Clear Image' by replacing the single Zustand state function with a module-level Set of listeners; registerImageReset now returns an unsubscribe function, the block extension uses it in its dispose proxy to prevent stale callback accumulation, and reset() clears the Set. Comprehensive Vitest coverage is included. Merging via rebase.

ivantha:

alikhere:
Hi @ivantha, Rebased onto latest main,  PTAL

ivantha:

PR #304, opened 2026-04-14
fix(geometric): make AffineImage tx/ty configurable via params

### LINKED ISSUE #300: [Bug] AffineImage operator ignores user parameters — translation values are hardcoded
**Describe the bug**
The `AffineImage` operator has hardcoded translation values (`tx=50, ty=100`) and never reads from `self.params`. Users cannot configure the transformation at all — every run produces the same fixed translation.

**To Reproduce**
1. Add an `AffineImage` operator to the pipeline
2. Set any custom translation or transform parameters
3. Run the pipeline — output always applies tx=50, ty=100 regardless of input

**Expected behavior**
The operator should read configurable parameters (`tx`, `ty`, angle, scale) from `self.params` instead of using hardcoded values.

**Screenshots**
N/A

**Environment**
- Python version: 3.12

**Additional context**
Relevant code in `app/operators/geometric/affine_image.py`:
```python

ajithhraj:
**Describe the bug**
`AffineImage` operator had hardcoded translation values (`tx=50, ty=100`) and never read from `self.params`. Users could not configure the transformation at all.

Fixes #300

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Manual testing

Verified that setting custom tx/ty values in the block correctly translates the image by the specified pixel offsets.

## Checklist
- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/geometric/affine_image.py
+++ b/imagelab-backend/app/operators/geometric/affine_image.py
@@ -7,5 +7,7 @@ from app.operators.base import BaseOperator
 class AffineImage(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         rows, cols = image.shape[:2]
-        M = np.float64([[1, 0, 50], [0, 1, 100]])
+        tx = float(self.params.get("tx", 50))
+        ty = float(self.params.get("ty", 100))
+        M = np.float64([[1, 0, tx], [0, 1, ty]])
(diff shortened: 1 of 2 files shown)


ivantha:
Thanks @ajithhraj! Closing this as #82 addresses the same AffineImage hardcoded-params bug by exposing all 12 affine matrix coordinates, which is a more complete solution. Your fix is clean — please check #82 and see if your tx/ty approach would complement it.

PR #309, opened 2026-04-17
Added restore option to restore default parameters

### LINKED ISSUE #308: [Feature] Added restore option to restore default parameters
Yes. There is no convenient, per-block way to reset a block’s parameters to their original defaults. Users who tweak block fields (numbers, dropdowns, colours, etc.) must manually revert changes or delete & re-add the block, which is slow and error-prone when iterating pipelines or demonstrating examples.

Add a per-block “Restore defaults” context-menu item that resets that block’s editable fields to the defaults defined in its JSON definition.

Workspace-level “Reset selected block” toolbar button (less discoverable than right-click).
A small inline reset icon on the block UI (clutters the block surface).
Per-block explicit “reset” field in JSON/extensions (requires extra manual work for each block).
Re-creating blocks on reset (current fallback, but user-unfriendly).
Storing runtime defaults on the block prototype rather than extracting from JSON (works but duplicates data and diverges from the single-source-of-truth JSON definitions).

chetanbarakki:
## Description
When experimenting with block parameters in the Blockly workspace, it's easy to change values and lose the original defaults
This makes reproducing examples harder, quick experimentation slower (users must recreate blocks to reset), sharing pipelines risk exposing accidental parameter changes.

Fixes #308 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

https://github.com/user-attachments/assets/a247e2a6-9545-4f06-8078-1978f388eae0

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-frontend/src/blocks/definitions/index.ts
+++ b/imagelab-frontend/src/blocks/definitions/index.ts
@@ -34,10 +34,85 @@ function registerOddKernelValidator() {
   });
 }
 
+/**
+ * Store default values per block type for restore-to-defaults functionality.
+ * Maps blockType -> fieldName -> defaultValue
+ */
+const blockDefaults = new Map<string, Record<string, unknown>>();
+
+type ContextMenuItem = {
+  enabled?: boolean;
+  text: string;
+  callback?: (...args: unknown[]) => void;
+};
+
+/**
+ * Extract default field values from a block's JSON definition.
+ */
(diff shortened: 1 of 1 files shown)


chetanbarakki:
@ivantha can you please review and say if there are any changes that has to be done

PR #321, opened 2026-04-23
fix(conversions): convert BGR/BGRA to gray before thresholding in GrayToBinary

### LINKED ISSUE #319: [Bug] GrayToBinary silently produces wrong output when given a color (BGR/BGRA) image
## Description

## Steps to Reproduce

```python
import cv2, numpy as np

```

## Expected Behaviour

`GrayToBinary` should either:

## Actual Behaviour

## Context

Jayant-kernel:
## Description

Fixes `GrayToBinary` silently producing a 3-channel output when given a colour (BGR or BGRA) image instead of a grayscale one.

Previously, `cv2.threshold` ran on each colour channel independently, producing a 3-channel result that was incorrectly labelled as binary. Now the operator converts BGR/BGRA input to grayscale first — consistent with how `ColorToBinary` handles the same case.

Fixes #319

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] New tests added — `tests/operators/conversions/test_gray_to_binary.py`
  - grayscale input works as before
  - BGR input is converted to gray before thresholding (single-channel output)
  - BGRA input is converted to gray before thresholding (single-channel output)
  - default params produce correct binary output
- [x] Existing tests pass

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/conversions/gray_to_binary.py
+++ b/imagelab-backend/app/operators/conversions/gray_to_binary.py
@@ -6,6 +6,12 @@ from app.operators.base import BaseOperator
 
 class GrayToBinary(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
+        if image.ndim == 3:
+            if image.shape[2] == 4:
+                image = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
+            else:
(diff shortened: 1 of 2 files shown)


ivantha:
Closing as stale during repository cleanup. Thanks for the contribution. Feel free to reopen against the latest main if you'd like this revisited.

Issue #302, opened 2026-04-14
[Bug] ScaleImage crashes when fx or fy is zero or negative


ajithhraj:
**Describe the bug**
`ScaleImage` passes `fx` and `fy` directly to `cv2.resize()` without any validation. Zero or negative scale factors cause an OpenCV crash.

**To Reproduce**
1. Add a `ScaleImage` operator to the pipeline
2. Set `fx=0` or `fy=0` (or any negative value)
3. Run the pipeline — crashes with `cv2.error`

**Expected behavior**
Should clamp `fx` and `fy` to a minimum positive value (e.g. `max(0.01, fx)`) before passing to OpenCV, consistent with how other operators validate their parameters.

**Screenshots**
N/A

**Environment**
- Python version: 3.12

**Additional context**
Relevant code in `app/operators/geometric/scale_image.py`:
```python
fx = float(self.params.get("fx", 1))
fy = float(self.params.get("fy", 1))
# no validation — zero/negative values crash cv2.resize
return cv2.resize(image, dsize=(0, 0), fx=fx, fy=fy, ...)
```

ivantha:
Closing: this is already fixed on `main`.

ScaleImage now rejects non-finite and non-positive fx/fy with a ValueError before touching cv2.resize, so the crash is gone (the remedy is explicit validation rather than the clamping the issue suggested).

Evidence: imagelab-backend/app/operators/geometric/scale_image.py:25-31 (commit 577b728 'fix(geometric): validate ScaleImage factors and prevent zero-scale crash').

Thanks for the report. If you still see this on `main`, comment here and we will reopen.

PR #322, opened 2026-04-23
fix(geometric): validate RotateImage scale is positive and finite

### LINKED ISSUE #320: [Bug] RotateImage scale=0 or negative silently corrupts the image with no validation error
## Description

`RotateImage` accepts any numeric value for its `scale` parameter but performs **no validation**. A `scale` of `0` or a negative value silently produces a corrupted output image instead of raising a clear error.

## Steps to Reproduce

```python
import cv2, numpy as np
from app.operators.geometric.rotate_image import RotateImage

img = np.ones((100, 100, 3), dtype=np.uint8) * 128

# scale = 0: every output pixel is mapped to the centre of the source image
out_zero = RotateImage({"angle": 45, "scale": 0}).compute(img)
print("scale=0 — all pixels same value:", np.unique(out_zero))
# e.g. array([128]) — entire image collapses to one colour, no error raised

# scale = -1: image is inverted/mirrored in an unexpected way
out_neg = RotateImage({"angle": 45, "scale": -1}).compute(img)
print("scale=-1 — output differs silently:", out_neg[0, 0])
```

**What cv2.getRotationMatrix2D does with scale=0:**
The resulting matrix maps *every* destination pixel back to the image centre, so the entire output becomes a uniform colour equal to the centre pixel — the image is silently destroyed.

## Expected Behaviour

`RotateImage` should validate that `scale > 0` and raise a `ValueError` with a descriptive message when an invalid value is provided, consistent with the fix applied to `ScaleImage` in issue #302.

## Actual Behaviour

No error is raised. A `scale` of `0` produces a uniform solid-colour image; a negative `scale` produces an unexpected mirrored/distorted result. Both cases silently corrupt the pipeline output.

## Context

- `ScaleImage` had an identical class of bug (issue #302) and was fixed by adding explicit validation for `fx > 0` and `fy > 0`.
- `RotateImage.scale` uses the same OpenCV `getRotationMatrix2D` API and needs the same guard.

Jayant-kernel:
## Description

Fixes `RotateImage` silently corrupting the output when `scale` is zero, negative, or non-finite.

With `scale=0`, `cv2.getRotationMatrix2D` maps every destination pixel to the image centre, producing a uniform solid-colour output with no error raised. Negative and non-finite values produce similarly unexpected results.

This adds the same class of validation already applied to `ScaleImage` (issue #302).

Fixes #320

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

- [x] New tests added — `tests/operators/geometric/test_rotate_image.py`
  - default and valid params work correctly
  - `scale=0`, negative, `inf`, `-inf`, `nan` all raise `ValueError`
  - shape is preserved for valid inputs
- [x] Existing tests pass

## Checklist

- [x] Tests pass locally

### DIFF
--- a/imagelab-backend/app/operators/conversions/gray_to_binary.py
+++ b/imagelab-backend/app/operators/conversions/gray_to_binary.py
@@ -6,6 +6,12 @@ from app.operators.base import BaseOperator
 
 class GrayToBinary(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
+        if image.ndim == 3:
+            if image.shape[2] == 4:
+                image = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
+            else:
+                image = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+
         threshold_value = float(self.params.get("thresholdValue", 0))
         max_value = float(self.params.get("maxValue", 255))
         _, dst = cv2.threshold(image, threshold_value, max_value, cv2.THRESH_BINARY)
--- a/imagelab-backend/app/operators/geometric/rotate_image.py
+++ b/imagelab-backend/app/operators/geometric/rotate_image.py
@@ -1,3 +1,5 @@
+import math
+
 import cv2
 import numpy as np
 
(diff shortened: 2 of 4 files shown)


ivantha:
Closing as stale during repository cleanup. Thanks for the contribution. Feel free to reopen against the latest main if you'd like this revisited.

PR #345, opened 2026-08-03
fix: normalise image format from mime type across all upload paths

### LINKED ISSUE #344: [Bug] Uploading GIF or SVG files causes Unsupported image format error
When a user uploads a GIF or SVG file, the pipeline fails with a backend error. The frontend extracts the image format from the browser mime type using a raw string split, which produces invalid values:

- image/gif gives gif which the backend does not support
- image/svg+xml gives svg+xml which the backend does not support

The error appears in the pipeline results panel:

`
Error in Step 2 (imageconvertions_grayimage):
ValueError: Unsupported image format: 'gif'.
Supported: ['bmp', 'jpeg', 'jpg', 'png', 'tif', 'tiff', 'webp']
`

## To Reproduce

1. Open the app at http://localhost:3100
2. Use the Read Image block to upload a GIF file (e.g. sample_640x426.gif)
3. Add any operator block (e.g. Gray the image)
4. Click Run
5. See the error in the bottom right of the screen

## Expected behavior

The app should gracefully handle unsupported mime types by falling back to PNG, rather than crashing with a backend error.

## Screenshots

Image

## Root cause

The utility getImageFormatFromMimeType in src/utils/imageData.ts already exists and is used correctly in CameraCaptureModal.tsx, but two other upload paths bypass it:

`	s
// src/blocks/extensions/readImageExtension.ts
// src/components/Preview/PreviewPane.tsx
const format = file.type.split('/')[1] || 'png';  // broken for image/svg+xml, image/gif
`

The fix is to update getImageFormatFromMimeType to validate against the supported format list and fall back to png for anything unsupported, then use it consistently across all upload paths.

shree1071:
## Summary

Fixes #344

Uploading a GIF or SVG file caused a backend crash because the frontend was extracting the image format with a raw mime type split:

`	s
const format = file.type.split('/')[1] || 'png';
// image/svg+xml -> 'svg+xml'  (backend error)
// image/gif     -> 'gif'       (backend error)
`

A utility function getImageFormatFromMimeType already existed and was used correctly in CameraCaptureModal.tsx. Two other upload paths were bypassing it.

## Changes

- src/utils/imageData.ts — updated getImageFormatFromMimeType to validate against the backend's supported format list and fall back to png for anything unsupported or compound
- src/blocks/extensions/readImageExtension.ts — replaced raw split with getImageFormatFromMimeType
- src/components/Preview/PreviewPane.tsx — replaced raw split with getImageFormatFromMimeType
- 	ests/utils/imageData.test.ts — 6 new test cases (svg+xml, gif, empty string, webp, tiff, empty mime)

## Testing

All 68 frontend tests pass.

| Mime type | Before | After |
|-----------|--------|-------|
| image/jpeg | jpeg | jpeg |
| image/png | png | png |
| image/svg+xml | svg+xml (backend error) | png |
| image/gif | gif (backend error) | png |
| empty | png | png |

##AFTER
image

### DIFF
--- a/imagelab-frontend/src/blocks/extensions/readImageExtension.ts
+++ b/imagelab-frontend/src/blocks/extensions/readImageExtension.ts
@@ -1,4 +1,5 @@
 import * as Blockly from "blockly";
+import { getImageFormatFromMimeType } from "../../utils/imageData";
 import { usePipelineStore } from "../../store/pipelineStore";
 
 function setFilenameLabel(block: Blockly.Block, value: string) {
@@ -21,7 +22,7 @@ function initReadImageBlock(block: Blockly.Block) {
     const file = fileInput.files?.[0];
     if (!file) return;
 
-    const format = file.type.split("/")[1] || "png";
+    const format = getImageFormatFromMimeType(file.type);
     const reader = new FileReader();
     reader.onload = () => {
       const dataUrl = reader.result as string;
--- a/imagelab-frontend/src/components/Preview/PreviewPane.tsx
+++ b/imagelab-frontend/src/components/Preview/PreviewPane.tsx
@@ -1,6 +1,7 @@
 import { useRef, useState } from "react";
 import { Camera, Image, ImageDown, Timer, Trash2, Upload, ZoomIn, ZoomOut } from "lucide-react";
 import { usePipelineStore } from "../../store/pipelineStore";
+import { getImageFormatFromMimeType } from "../../utils/imageData";
 import ImageDisplay from "./ImageDisplay";
 
 function ZoomControls({
@@ -72,7 +73,7 @@ export default function PreviewPane() {
       const [, base64 = ""] = dataUrl.split(",", 2);
       if (!base64) return;
 
-      const format = file.type.split("/")[1] || "png";
+      const format = getImageFormatFromMimeType(file.type);
       setOriginalImage(base64, format, file.name);
     };
     reader.readAsDataURL(file);
--- a/imagelab-frontend/src/utils/imageData.ts
+++ b/imagelab-frontend/src/utils/imageData.ts
@@ -26,9 +26,16 @@ export function blobToDataUrl(blob: Blob): Promise<string> {
   });
 }
 
+const SUPPORTED_IMAGE_FORMATS = new Set(["png", "jpg", "jpeg", "webp", "bmp", "tiff", "tif"]);
+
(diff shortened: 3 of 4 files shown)


ivantha:
### Automated review summary

**Triage:** bug  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/quality]** `imagelab-frontend/tests/utils/imageData.test.ts:98` This comment (and the PR description's before/after table) attributes the SVG fix to the backend's supported-format whitelist, but the backend's decode_base64_image() never receives image_format at all — it decodes purely from raw byte content via cv2.imdecode, and only uses image_format later when re-encoding an already-decoded image. A raw SVG (vector XML) upload fails at that decode step regardless of whether format is 'svg+xml' or 'png', so this change doesn't actually make SVG uploads succeed — it only changes the resulting error message from 'Unsupported image format' to 'Could not decode image data'. The fix is genuinely effective for formats that decode fine as raster but aren't in the backend's output-encoding whitelist (e.g. GIF), just not for SVG as the comment implies.
- **[minor/quality]** `imagelab-frontend/src/utils/imageData.ts:29` SUPPORTED_IMAGE_FORMATS duplicates the backend's SUPPORTED_FORMATS set (imagelab-backend/app/utils/image.py) with no shared source of truth. The two lists match exactly today, but nothing enforces that going forward — a future change to either whitelist alone (e.g. the backend adding avif support) will silently desync the two and reintroduce the same class of bug this PR fixes.
- **[nit/test]** `imagelab-frontend/tests/utils/imageData.test.ts:84` The new tests exercise jpeg, webp, tiff, svg+xml and gif, but two formats newly added to SUPPORTED_IMAGE_FORMATS — 'bmp' and 'jpg' — have no test anywhere in this file confirming they pass through unchanged rather than falling back to png.

**Conventions**

This is a small, well-targeted frontend fix: it routes two upload paths (readImageExtension.ts, PreviewPane.tsx) through the existing getImageFormatFromMimeType utility instead of a raw mime-type split, and tightens that utility to fall back to png for compound or unsupported subtypes, backed by six new unit tests. The new allow-list happens to mirror the backend's supported-format whitelist exactly but isn't derived from it, so the two can silently drift apart over time. The SVG case specifically isn't fully fixed by this change — the backend's image decoder never inspects the declared format, so SVG uploads still fail after this PR, just with a different error message than before. Otherwise the diff is minimal, self-contained, and well tested; safe to merge once the SVG claim in the description/tests is corrected or acknowledged as a partial fix.

PR #346, opened 2026-08-05
Added macro crud endpoints and UI for macros

chetanbarakki:
## Description

- Added REST endpoints (GET, POST, PUT, DELETE /api/v1/macros) protected by graph validation and dependency safety checks (blocking deletion if a macro is actively used). 
- Added The Additive Graph Expansion Service that recursively unrolls macro references into a flat array of steps right before execution, seamlessly passing it to the existing PipelineExecutor. 
- Zustand/React store and API clients to interact with the new backend macro endpoints.  
- The extraction utility (extractMacroGraph.ts) that walks Blockly connections (nextConnection, inputs) of drag-selected blocks and maps them into the backend-compliant JSON graph format.  
- The "Create Macro" action modal in the workspace toolbar, letting users choose which parameters to expose and handle validation errors gracefully

Fixes #39 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/alembic/versions/c1a2b3c4d5e6_add_is_macro_to_pipeline.py
+++ b/imagelab-backend/alembic/versions/c1a2b3c4d5e6_add_is_macro_to_pipeline.py
@@ -1,0 +1,30 @@
+"""add is_macro to pipeline table
+
+Revision ID: c1a2b3c4d5e6
+Revises: f78ce0d472f1
+Create Date: 2026-07-28 22:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "c1a2b3c4d5e6"
+down_revision: str | None = "f78ce0d472f1"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "pipeline",
+        sa.Column("is_macro", sa.Boolean(), server_default=sa.text("false"), nullable=False),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("pipeline", "is_macro")
--- a/imagelab-backend/app/main.py
+++ b/imagelab-backend/app/main.py
@@ -7,7 +7,7 @@ from fastapi.middleware.cors import CORSMiddleware
 from app.config import get_settings
 from app.exceptions import register_exception_handlers
 from app.middleware.rate_limit import ShareRateLimitMiddleware
-from app.routers import batch, persistence, pipeline
+from app.routers import batch, macros, persistence, pipeline
 
 logger = logging.getLogger(__name__)
 
@@ -49,6 +49,7 @@ app.include_router(pipeline.router, prefix="/api")
 app.include_router(persistence.router, prefix="/api")
 app.include_router(persistence.share_router, prefix="/api")
 app.include_router(batch.router, prefix="/api")
+app.include_router(macros.router, prefix="/api")
 
 if __name__ == "__main__":
     import uvicorn
--- a/imagelab-backend/app/models/persistence.py
+++ b/imagelab-backend/app/models/persistence.py
@@ -12,8 +12,7 @@ class Pipeline(SQLModel, table=True):
     id: uuid.UUID = Field(default_factory=uuid.uuid4, primary_key=True, index=True)
     name: str
     owner_id: str | None = Field(default=None, nullable=True)
-    created_at: datetime = Field(default_factory=datetime.now(UTC))
-    updated_at: datetime = Field(default_factory=datetime.now(UTC))
+    is_macro: bool = Field(default=False, nullable=False)
     created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
     updated_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
 
--- a/imagelab-backend/app/routers/macros.py
+++ b/imagelab-backend/app/routers/macros.py
@@ -13,6 +13,7 @@ from typing import Annotated, Any
 
 from fastapi import APIRouter, Depends, HTTPException, status
 from pydantic import BaseModel
+from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 from sqlmodel import Session, select
 
@@ -81,10 +82,26 @@ def _get_macro_or_404(session: Session, macro_id: uuid.UUID) -> Pipeline:
 
 
 def _check_macro_in_use(session: Session, macro_id: uuid.UUID) -> bool:
-    """Return True if macro_id is referenced in any existing pipeline or macro's PipelineVersion."""
+    """Return True if macro_id is referenced in any existing pipeline or macro's latest PipelineVersion."""
     macro_id_str = str(macro_id)
-    versions = session.exec(select(PipelineVersion)).all()
-    for ver in versions:
+
+    # Subquery to get the latest version for each pipeline
+    latest_versions_subquery = (
+        select(PipelineVersion.pipeline_id, func.max(PipelineVersion.version_number).label("max_version"))
+        .group_by(PipelineVersion.pipeline_id)
+        .subquery()
+    )
+
+    # Query only the latest versions
+    latest_versions = session.exec(
+        select(PipelineVersion).join(
+            latest_versions_subquery,
+            (PipelineVersion.pipeline_id == latest_versions_subquery.c.pipeline_id)
+            & (PipelineVersion.version_number == latest_versions_subquery.c.max_version),
+        )
+    ).all()
+
+    for ver in latest_versions:
         if ver.pipeline_id == macro_id:
             continue
         pipeline = session.get(Pipeline, ver.pipeline_id)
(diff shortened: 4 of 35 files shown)


ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-revision

**Findings**
- **[major/bug]** `imagelab-frontend/src/blocks/macroBlock.ts:26` registerMacroBlock() returns early whenever `macro_${macro.id}` already exists as a key in Blockly.Blocks. Since a macro's id never changes across edits, updating a macro's name or exposed parameters through the new PUT endpoint never updates its already-registered sidebar block for the rest of the session; the block keeps rendering the stale name/fields until the page is hard-refreshed.
- **[major/bug]** `imagelab-backend/app/routers/macros.py:86` _check_macro_in_use() scans every PipelineVersion row for every pipeline instead of just each pipeline's latest version (the pattern get_macro/update_macro both use elsewhere in this same file). Once any now-superseded version of some pipeline ever referenced a macro, that macro can never be deleted again even after all current pipelines have stopped using it, which contradicts the module's own docstring claim of blocking deletion only when 'actively referenced'.
- **[major/bug]** `imagelab-backend/app/services/batch_executor.py:51` If expand_macro_steps() raises for any reason (macro deleted, DB error, cycle), this except block only logs a warning and leaves exec_pipeline as the original, unexpanded pipeline, which still contains unresolved macro steps. execute_pipeline() then runs anyway, so a batch job fails per image with a confusing 'unknown step type' style error instead of surfacing the real macro-expansion failure.
- **[minor/security]** `imagelab-backend/app/routers/macros.py:173` create_macro and update_macro (line 258) both return detail=f"Failed to create/update macro: {exc}", embedding the raw caught exception text in the client-facing 500 response. Every other handler in the sibling persistence.py router returns a generic message like 'Internal error' instead, so this leaks internal exception detail and is inconsistent with the rest of the codebase.
- **[minor/quality]** `imagelab-frontend/src/store/useMacroStore.ts:51` loadMacros() fetches the macro list and then issues one additional getMacro() request per macro via Promise.all, an N+1 network pattern that reruns on every Sidebar mount and after every create/edit/delete. list_macros already omits the graph payload, so returning it in the list response (or adding a batch-fetch endpoint) would avoid this as the saved-macro library grows.
- **[minor/quality]** `imagelab-frontend/src/utils/expandPipelineMacros.ts:56` The while loop that expands macro_ wrapper nodes has no visited-set or depth guard against a macro that directly or indirectly references itself; it relies entirely on the backend's create/update-time cycle check keeping useMacroStore's cached macros acyclic. Any gap in that guarantee would hang the browser tab in this loop with no recovery but a forced reload.

**Conventions**
- **PR title:** Descriptive, past-tense title rather than Conventional Commits style, e.g. 'feat: add macro CRUD endpoints and UI' instead of 'Added macro crud endpoints and UI for macros'.
- **Commits (2):** Neither commit message uses a Conventional Commits prefix or imperative mood ('Added macrocrud api and frontend', 'Fixed errors and added multi select blocks...'); the second reads like a fixup that should have been squashed into the first.
- _Retitle the PR and commits with Conventional Commits prefixes (feat:, fix:), squash the fixup commit, and fill in the PR template's type-of-change and testing checklists, which are currently all unchecked._

This PR adds a full macro (composite operator) feature: backend CRUD endpoints with graph cycle and type validation plus an in-use deletion guard, a graph-expansion service that unrolls macro references before execution, and a matching frontend with a Zustand store, dynamic Blockly block registration, 2-click range selection, and a Create Macro modal, backed by tests on both sides. The most impactful issue is that editing a macro's name or exposed parameters never updates its already-rendered sidebar block for the rest of the session, because registerMacroBlock() skips re-registration once a block type exists, so the PUT endpoint's effect is invisible until a hard refresh. Two more correctness gaps are worth fixing before merge: the delete guard in _check_macro_in_use checks every historical pipeline version instead of just the current one, making a macro permanently undeletable once any old, superseded version referenced it, and a failed macro expansion in batch processing is only logged as a warning, so the job silently continues with the unexpanded pipeline instead of surfacing a clear error. There is also a smaller information-disclosure smell in the two 500-error handlers and an N+1 network pattern when loading the macro list, both easy follow-ups.

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-revision
**GSoC scope:** overlaps reserved GSoC scope (issue #39, full)

**Findings**
- **[blocking/bug]** `imagelab-backend/app/routers/macros.py:114` _check_macro_in_use (the deletion 'dependency safety check' the PR description highlights) only matches a node whose type is the literal string 'macro_ref' inside pipeline_json['nodes']. Neither real shape the app actually persists matches that: a macro nested inside another saved macro is stored by extractMacroGraph.ts/registerMacroBlock as {type: 'macro_<childId>', ...} (imagelab-frontend/src/utils/extractMacroGraph.ts:249, imagelab-frontend/src/blocks/macroBlock.ts:73), and a regular pipeline that references a macro is stored by the pre-existing SavePipelineModal.tsx as pipeline_json={'steps': [...]}, so line 111's p_json.get('nodes', []) finds nothing at all for it. I verified both cases by running the PR's actual code with the FastAPI TestClient: a macro nested in another saved macro, and a macro referenced from a saved regular pipeline, both DELETE with 204 instead of the expected 409 -- the safety check never fires for any real usage pattern, only for the hand-crafted 'macro_ref' shape the PR's own test uses.
- **[major/bug]** `imagelab-frontend/src/hooks/useBlocklyWorkspace.ts:188` isDark was added to initWorkspace's useCallback dependency array, and the comment explaining why it had been deliberately excluded was deleted along with it ('adding isDark here would dispose and recreate the workspace on every toggle, losing user blocks'). Because the mount effect at line 197 depends on [initWorkspace], every isDark change now runs the cleanup (workspace.dispose()) and re-creates the whole Blockly canvas via Blockly.inject() again, even though the separate effect at lines 58-61 already swaps the theme live via workspace.setTheme() with no disposal at all. This reintroduces exactly the regression the deleted comment warned about, and does so on a prop (dark-mode toggle) users interact with far more often than readOnly, the prop this dispose/recreate pattern was previously scoped to.
- **[major/bug]** `imagelab-backend/app/services/graph_engine.py:66` create_macro/update_macro's validation (_validate_macro_payload -> validate_macro_graph) delegates cyclic-macro-nesting detection to expand_all_macros, which has the identical 'macro_ref'-only matching bug as the deletion-guard finding above. Since real nested-macro nodes are typed 'macro_<id>' and never literally 'macro_ref', this function's recursive active_macro_ids/GraphCycleError cycle guard (lines 77-80) never engages for macros built through the real UI -- macro A embedding macro B embedding macro A can be saved without a server-side error. The only protection is the client-side hasMacroCycle check in extractMacroGraph.ts, bypassable via a direct API call. The PR's own cycle test (test_macro_graph_validation_rejects_cycles) only covers a direct block-to-block cycle within one graph, not cross-macro nesting, so this gap isn't caught by CI.
- **[major/bug]** `imagelab-backend/app/services/batch_executor.py:49` The macro-detection gate here, and expand_macro_steps (graph_engine.py:305-312) that it calls, use the same 'macro_ref'/params.macro_id convention that real macro-block steps never have (extractPipeline in usePipeline.ts:9-40 emits {type: 'macro_<id>', params: {...}} for a macro block, with no macro_id key at all). Today this is masked because the frontend's extractExecutablePipeline already expands macros client-side before POSTing to /v1/batch-jobs, but this backend path exists specifically as that endpoint's own safety net (it accepts a raw pipeline form field from any caller, not just this frontend) and is not fit for that purpose: a real, unexpanded macro step would fail to be recognized, never get expanded, and be forwarded to execute_pipeline as-is, which has no handler for a 'macro_<uuid>' type and would fail every image in the batch.
- **[minor/test]** `imagelab-backend/tests/test_macro_api.py:157` test_macro_dependency_safety_check_blocks_deletion (and test_macro_expansion_and_step_inspection at line 200) hand-construct pipeline_json nodes as {'type': 'macro_ref', 'params': {'macro_id': ...}}. That is the shape _check_macro_in_use/expand_all_macros check for, but not the shape the real frontend ever produces ('macro_ref' does not appear anywhere in the frontend code, which always emits {'type': 'macro_<id>'}). Both tests pass while the actual integration is broken; a fixture built by going through create_macro with a genuinely nested macro block (matching what CreateMacroModal actually submits) would have caught the deletion-guard and cycle-validation findings above.
- **[minor/quality]** `imagelab-backend/app/routers/macros.py:145` list_macros issues one extra query per macro to fetch its latest PipelineVersion (N+1), and _check_macro_in_use (line 107) does a per-row session.get(Pipeline, ...) whose result is used only for a None-check and is otherwise discarded. Both loops could reuse the single joined/subquery pattern _check_macro_in_use itself already applies for its own version lookup a few lines above.
- **[minor/quality]** `imagelab-backend/app/routers/macros.py:311` delete_macro has no try/except around session.delete(macro)/session.commit(), unlike create_macro and update_macro in the same file, which both wrap their commits and translate IntegrityError/generic Exception into a clean HTTPException. Any failure here would surface as a raw unhandled 500 instead of following this file's own error-handling convention.
- **[nit/quality]** `imagelab-backend/app/routers/macros.py:51` MacroOut is defined but never used as a response_model or referenced anywhere else in the backend (imagelab-frontend/src/types/macro.ts:41's MacroItem is similarly unused on the frontend). Both look like leftovers from an earlier response-shape iteration and can be deleted.

**Conventions**
- **PR title:** Not Conventional-Commits style (no feat: prefix), uses past tense ('Added'), and is redundantly phrased -- 'macro crud endpoints and UI for macros' repeats 'macro(s)' at both ends.
- **Commits (8):** Only 2 of 8 titles use a Conventional Commits type prefix; the rest are plain descriptive titles, one has a typo ('detetion' for 'deletion'), and one bundles multiple unrelated changes into a single comma-separated message ('feature added batch expansion safeguards,fixed pr review errors and c...') rather than being split into atomic commits.
- _Retitle the PR and tighten commit messages to this repo's Conventional Commits convention, and check off the PR template's testing/checklist boxes since substantial new backend and frontend tests were in fact added._

This PR adds a complete macro (reusable composite-operator) feature: backend CRUD, graph/cycle validation, a deletion "dependency safety check", macro expansion at both interactive-run and batch-processing time, and a full Blockly sidebar/modal UI for creating, editing, and deleting macros. However, running the PR's own code confirms its two advertised server-side safety checks don't work: deleting a macro nested inside another saved macro, or referenced by a saved pipeline, both succeed instead of being blocked with a 409, because the checks look for a node shape the real frontend never produces. The same node-shape mismatch leaves the batch-processing macro-expansion path effectively dead code, currently masked only because the frontend already expands macros client-side before submitting a batch job. Separately, useBlocklyWorkspace.ts now adds isDark to initWorkspace's dependency array and deletes the comment explaining why it was deliberately excluded, reintroducing exactly the regression that comment warned about: toggling dark mode now disposes and rebuilds the entire Blockly canvas instead of using the existing, non-disruptive theme-swap effect a few lines above it.

PR #349, opened 2026-08-11
fix: convert BGR/BGRA to grayscale in GrayToBinary before thresholding

### LINKED ISSUE #319: [Bug] GrayToBinary silently produces wrong output when given a color (BGR/BGRA) image

shree1071:
## Summary

Fixes #319

The GrayToBinary operator is designed to threshold a grayscale image into a binary (black and white) image. When a color (BGR or BGRA) image is passed in, it applies cv2.threshold() directly on all channels and returns a 3 or 4-channel binary output instead of a proper 2D grayscale binary image.

## What was wrong

GrayToBinary.compute() passed the image directly to cv2.threshold() with no channel check. OpenCV happily thresholds multi-channel images per-channel, so the operator appeared to work but produced a result with the wrong number of channels — defeating the purpose of the operator.

## To Reproduce

1. Upload any color image
2. Add a **Conversions > Gray to Binary** block (without first converting to gray)
3. Click Run
4. The output has 3 channels instead of 1 — downstream operators expecting a grayscale binary image will behave incorrectly

## Fix

Added channel detection at the start of compute() that converts BGR and BGRA inputs to grayscale before thresholding, matching the operator's documented intent.

`python
if image.ndim == 3:
    if image.shape[2] == 4:
        image = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
    else:
        image = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
`

## Tests

2 new tests added to TestGrayToBinary in 	ests/test_conversion_operators.py:

- 	est_color_bgr_input_produces_grayscale_output — BGR input gives 2D output
- 	est_color_bgra_input_produces_grayscale_output — BGRA input gives 2D output

All 7 GrayToBinary tests pass. Full suite: 671 passed.

### DIFF
--- a/imagelab-backend/app/operators/conversions/gray_to_binary.py
+++ b/imagelab-backend/app/operators/conversions/gray_to_binary.py
@@ -8,5 +8,15 @@ class GrayToBinary(BaseOperator):
     def compute(self, image: np.ndarray) -> np.ndarray:
         threshold_value = float(self.params.get("thresholdValue", 127))
         max_value = float(self.params.get("maxValue", 255))
+
+        # Convert color images to grayscale before thresholding.
+        # Passing a BGR or BGRA image directly to cv2.threshold produces a
+        # multi-channel binary output instead of a proper grayscale binary image.
+        if image.ndim == 3:
+            if image.shape[2] == 4:
+                image = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
+            else:
+                image = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+
         _, dst = cv2.threshold(image, threshold_value, max_value, cv2.THRESH_BINARY)
         return dst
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Triage:** bug  |  **Recommendation:** ready-for-maintainer

**Findings**
- None flagged.

**Conventions**

GrayToBinary previously ran cv2.threshold directly on whatever image it was given, so BGR/BGRA input produced a 3- or 4-channel 'binary' image instead of a proper 2D grayscale one. This PR adds a channel check that converts BGR/BGRA to grayscale before thresholding, using the exact same detection pattern already established in OtsuThreshold, AdaptiveThreshold, and ColorToBinary, so it's consistent with codebase convention rather than a novel approach. The two added tests are deterministic, target the BGR and BGRA regressions directly, and assert both shape and binary-value-set. The diff is small and self-contained, the description is thorough and links the originating bug report (#319), and no correctness, security, or style issues were found — this looks ready to merge.

PR #183, opened 2026-03-12
Fix ColorToBinary default maxValue causing all-black output

Jayant-kernel:
## Description

Fixes `ColorToBinary` default behavior where `maxValue=0` produced all-black output.

Changes included:
- Backend: set `ColorToBinary` default `maxValue` from `0` to `255`
- Frontend block default: set `imageconvertions_colortobinary.maxValue` from `0` to `255`
- Added regression test to ensure default params produce `{0, 255}` output

Fixes #182

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- Local runtime verification script confirms `ColorToBinary({})` now returns values `[0, 255]`.
- Added test: `test_default_params_produce_non_black_output` in `imagelab-backend/tests/test_conversion_operators.py`.
- Full pytest suite was not run in this environment because `pytest` is not installed in the project venv.

- [ ] Existing tests pass
- [x] New tests added
- [x] Manual testing

## Screenshots (if applicable)

N/A

## Checklist

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [ ] I have added/updated documentation as needed
- [x] My changes generate no new warnings
- [ ] Tests pass locally

PR #350, opened 2026-08-11
fix(conversions): default ColorToBinary maxValue to 255 (frontend + backend)

### LINKED ISSUE #182: [Bug] ColorToBinary default maxValue=0 causes all-black output
**Describe the bug**
`ColorToBinary` defaults `maxValue` to `0`, so thresholded pixels are written as `0` and output becomes fully black when default settings are used.

**To Reproduce**
1. Upload a color image.
2. Add `ColorToBinary` block with default parameters.
3. Run the pipeline.
4. Output image is completely black.

**Expected behavior**
Default behavior should produce a visible binary image. `maxValue` should default to `255`.

**Screenshots**
N/A

**Environment:**
- OS: Windows 11
- Browser: Chrome
- Node.js version: 18.x
- Python version: 3.12

**Additional context**
Locally validated in backend:
- `ColorToBinary({})` returns only `[0]`.
- `ColorToBinary({"thresholdValue":100,"maxValue":255})` returns `[0, 255]`.

Code paths currently using `maxValue=0` defaults:
- backend operator: `imagelab-backend/app/operators/conversions/color_to_binary.py`
- frontend block: `imagelab-frontend/src/blocks/definitions/conversions.blocks.ts`

### LINKED ISSUE #319: [Bug] GrayToBinary silently produces wrong output when given a color (BGR/BGRA) image

shree1071:
## Summary

Fixes #182: `ColorToBinary` defaults `maxValue` to `0`, producing a fully black output.

## Problem

`maxValue` was wrong in two places, and both had to change together.

**Backend** (`imagelab-backend/app/operators/conversions/color_to_binary.py`): `self.params.get("maxValue", 0)` fell back to `0` instead of `255`.

**Frontend** (`imagelab-frontend/src/blocks/definitions/conversions.blocks.ts`): the Blockly block declared `maxValue` with `value: 0`.

The frontend half is the one that actually bites. `extractPipeline` (`imagelab-frontend/src/hooks/usePipeline.ts`) serializes every named field's current value into the request unconditionally, so a freshly dragged block always sent `maxValue: 0` explicitly. Since `.get("maxValue", 255)` only falls back when the key is *absent*, fixing the backend alone would never have changed what the UI produces. The block definition is the de-facto default; the backend fallback only protects callers that hit the API without the UI.

With `maxValue = 0`, `cv2.threshold` writes `0` to every above-threshold pixel, so the output is entirely black and no error is surfaced.

## Changes

- `color_to_binary.py`: fallback `0` to `255`.
- `conversions.blocks.ts`: block field default `0` to `255`. This matches the sibling `GrayToBinary` block, which was already at `255`.
- New regression test `test_default_params_produce_non_black_output`, asserting default params yield a real `{0, 255}` split.
- Hardened the two existing `GrayToBinary` colour-input tests. Their pixel values (`128` and `200`) both sat above the default threshold of `127`, so the output collapsed to all-white and `issubset({0, 255})` passed without proving a binary split. Values now straddle the threshold and the assertion is an equality check.

## Verification

- Reproduced on `main`: `ColorToBinary({})` returns `[0]`.
- Mutation-checked the regression test: reverted the source fix, confirmed the test fails (`assert {np.uint8(0)} == {0, 255}`), restored it, confirmed it passes. The test cannot pass against the old default.
- Backend: 684 tests pass, `ruff check` and `ruff format --check` clean.
- Frontend: 93 tests pass, `tsc --noEmit` and `eslint` clean.

## How to reproduce the original bug

1. Open the app at http://localhost:3100 and upload any image.
2. Add a **Conversions > Color to Binary** block with default settings.
3. Click Run.

Before: the processed panel shows a completely black image, no error. After: a binary image.

## Notes

- Rebased onto `main` and squashed to one commit. The branch previously carried a `GrayToBinary` BGR/BGRA commit that duplicated #349, which had already merged and closed #319. That commit was byte-identical to `main` and was the source of the merge conflict, so it has been dropped. This PR no longer touches `gray_to_binary.py` and no longer relates to #319.
- Out of scope, filed separately: `thresholdValue` still defaults to `0`, so default params now yield a mostly-white image rather than a useful binary split. `thresholding_applythreshold` (`thresholding.blocks.ts:6`) has the identical `maxValue: 0` bug on a different block.

### DIFF
--- a/imagelab-backend/app/operators/conversions/color_to_binary.py
+++ b/imagelab-backend/app/operators/conversions/color_to_binary.py
@@ -14,7 +14,7 @@ class ColorToBinary(BaseOperator):
         threshold_type_name = self.params.get("thresholdType", "threshold_binary")
         threshold_type = THRESHOLD_TYPES.get(threshold_type_name, cv2.THRESH_BINARY)
         threshold_value = float(self.params.get("thresholdValue", 0))
-        max_value = float(self.params.get("maxValue", 0))
+        max_value = float(self.params.get("maxValue", 255))
 
         if len(image.shape) == 3 and image.shape[2] == 4:
             gray = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
--- a/imagelab-backend/tests/test_conversion_operators.py
+++ b/imagelab-backend/tests/test_conversion_operators.py
@@ -84,21 +84,23 @@ class TestGrayToBinary:
 
     def test_color_bgr_input_produces_grayscale_output(self):
         """BGR images should be converted to grayscale before thresholding — not produce 3-channel binary."""
-        color = np.full((100, 100, 3), 128, dtype=np.uint8)
-        color[50:, :] = [200, 200, 200]
+        # Values straddle the default threshold (127) so the split is observable
+        color = np.full((100, 100, 3), 50, dtype=np.uint8)  # below threshold -> 0
+        color[50:, :] = [200, 200, 200]  # above threshold -> 255
         result = GrayToBinary({}).compute(color)
         assert result.dtype == np.uint8
         assert result.ndim == 2, f"Expected 2D grayscale output for BGR input, got shape {result.shape}"
-        assert set(np.unique(result)).issubset({0, 255})
+        assert set(np.unique(result)) == {0, 255}
 
     def test_color_bgra_input_produces_grayscale_output(self):
         """BGRA images should be converted to grayscale before thresholding — not produce 4-channel binary."""
-        bgra = np.full((100, 100, 4), 128, dtype=np.uint8)
-        bgra[50:, :] = [200, 200, 200, 255]
+        # Values straddle the default threshold (127) so the split is observable
+        bgra = np.full((100, 100, 4), 50, dtype=np.uint8)  # below threshold -> 0
+        bgra[50:, :] = [200, 200, 200, 255]  # above threshold -> 255
         result = GrayToBinary({}).compute(bgra)
         assert result.dtype == np.uint8
         assert result.ndim == 2, f"Expected 2D grayscale output for BGRA input, got shape {result.shape}"
-        assert set(np.unique(result)).issubset({0, 255})
+        assert set(np.unique(result)) == {0, 255}
 
 
 # ColorToBinary
@@ -130,6 +132,14 @@ class TestColorToBinary:
     def test_output_is_uint8(self, color_image):
         result = ColorToBinary({"thresholdValue": 127, "maxValue": 255}).compute(color_image)
         assert result.dtype == np.uint8
+
+    def test_default_params_produce_non_black_output(self):
+        # Regression for #182: maxValue defaulted to 0, so every thresholded pixel
+        # was written as 0 and the whole output came back black.
+        img = np.zeros((100, 100, 3), dtype=np.uint8)
+        img[50:, :] = 200  # above the default threshold of 0 -> 255
+        result = ColorToBinary({}).compute(img)
+        assert set(np.unique(result)) == {0, 255}
 
 
 # ColorMaps
--- a/imagelab-frontend/src/blocks/definitions/conversions.blocks.ts
+++ b/imagelab-frontend/src/blocks/definitions/conversions.blocks.ts
@@ -120,7 +120,7 @@ export const conversionsBlocks = [
       },
       { type: "input_dummy" },
       { type: "field_number", name: "thresholdValue", value: 0, min: 0 },
-      { type: "field_number", name: "maxValue", value: 0, min: 0 },
+      { type: "field_number", name: "maxValue", value: 255, min: 0 },
     ],
     inputsInline: false,
     previousStatement: null,


ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `imagelab-frontend/src/blocks/definitions/conversions.blocks.ts:123` The ColorToBinary Blockly block still defines maxValue with a default of 0 (line 123), and this file is not touched by the diff. Toolbar.tsx's Run handler calls extractPipeline() (imagelab-frontend/src/hooks/usePipeline.ts, lines 15-21), which unconditionally reads every field's current value into the request params, so a freshly dragged ColorToBinary block still sends maxValue: 0 explicitly. Since self.params.get('maxValue', 255) only falls back to 255 when the key is absent, this backend fix never triggers for that request, and the exact repro steps in the PR description (drag the block with default settings, click Run) will likely still produce an all-black image. A prior PR (#183, closed unmerged after going stale) fixed both the backend and this frontend default together; this PR reintroduces only the backend half. _(independently verified during this review run: the block default is still `value: 0` at this head, `extractPipeline` sends every named field unconditionally, and the toolbox entry in `categories.ts` applies no field override, so `maxValue: 0` does reach the backend.)_
- **[minor/test]** `imagelab-backend/tests/test_conversion_operators.py:92` test_color_bgr_input_produces_grayscale_output (line 92) and test_color_bgra_input_produces_grayscale_output (line 101) use pixel values 128 and 200, both of which exceed the default threshold of 127, so the whole output collapses to a single value (verified by running it: np.unique returns only [255]). set(np.unique(result)).issubset({0, 255}) then passes trivially without confirming an actual black/white split. The sibling test_output_is_binary_values tests earlier in the same file avoid this by picking pixel values on both sides of an explicit threshold and asserting equality with {0, 255} rather than subset.

**Conventions**
- **PR title:** Names only the ColorToBinary fix; a second commit fixes an unrelated GrayToBinary bug (closes #319) that the title never mentions.
- **Description:** Covers only issue #182 and color_to_binary.py; the GrayToBinary grayscale-conversion fix and its tests are never described, and #319 (which this PR actually closes) is not linked.
- **Commits (2):** clean, conventional history
- _Both fixes are commit-separated cleanly, but the title and description surface only the ColorToBinary half, so update the description to disclose and link both fixes, or split into two PRs._

The ColorToBinary backend fix itself is correct and matches issue #182's root cause, but it is incomplete: the Blockly block definition (imagelab-frontend/src/blocks/definitions/conversions.blocks.ts, line 123) still defaults maxValue to 0, and since the frontend always sends every field's current value explicitly, the exact reproduction steps in this PR's own description will likely still produce an all-black image after merge. The bundled GrayToBinary commit is a separate, verified fix for a distinct, currently open bug (#319), but neither the title nor the description discloses or links it. The two new GrayToBinary regression tests are also weaker than the existing convention in the same file: their chosen pixel values both land above the default threshold, so the assertions can't actually distinguish a real binary split from an all-white output. Recommend asking the author to add the frontend default change before merging, and to update the description to cover both fixes.

ivantha:
### Automated review summary

**Findings**
- **[major/quality]** `imagelab-backend/app/operators/conversions/gray_to_binary.py:15` This commit duplicates PR #349 (same author, titled identically "fix: convert BGR/BGRA to grayscale in GrayToBinary before thresholding"), which was already merged to main at 2026-08-11T08:51:54Z closing issue #319, 26 minutes before this PR (#350) was opened. The branch was never rebased onto that merge, so this hunk is redundant against current main and will produce a no-op/conflict rather than new value when merged.
- **[minor/test]** `imagelab-backend/tests/test_conversion_operators.py:85` test_color_bgr_input_produces_grayscale_output and test_color_bgra_input_produces_grayscale_output duplicate method names already present in main's test_conversion_operators.py (added alongside the already-merged PR #349), with only the pixel values changed (straddling vs. uniform), so these collide rather than add new coverage once rebased.
- **[minor/bug]** `imagelab-backend/app/operators/conversions/gray_to_binary.py:18` The new channel check only branches on shape[2]==4 vs. else (treated as BGR); it doesn't handle a 3-D single-channel (H, W, 1) input the way the sibling GrayImage operator explicitly does (squeeze to 2-D). A (H, W, 1) array reaching this branch falls into the BGR2GRAY call and cv2 will raise, since BGR2GRAY requires a 3-channel source.

**Conventions**
- **Commits (3):** First commit (22f6480) duplicates PR #349 (same author, same message, already merged to main 26 minutes before this PR was opened, closing #319), so the branch needs a rebase to drop that redundant commit and its duplicate-named tests.
- _Rebase onto latest main before merge; once the redundant GrayToBinary commit and its duplicate tests are dropped, only the still-needed ColorToBinary maxValue fix remains._

This PR bundles two bug fixes but only one is still needed: the ColorToBinary maxValue default fix (backend color_to_binary.py:17 and frontend conversions.blocks.ts:123) is correct, minimal, and closes the still-open #182. The GrayToBinary BGR/BGRA-to-grayscale commit and its two new tests, however, duplicate PR #349 (merged by the same author 26 minutes before this PR was opened, already closing #319), so that hunk is dead weight against current main and its tests collide by name with ones already in the suite. Ask the author to rebase onto main to drop the redundant commit before this can merge cleanly; once rebased, the remaining channel-detection logic could also use a defensive branch for the documented (H, W, 1) single-channel case, matching the pattern GrayImage already follows.

ivantha:
@shree1071 heads up: I used maintainer edit access to rebase and squash this branch, so **your local copy has diverged**. Before you do any more work on it:

```bash
git fetch origin
git checkout fix/color-to-binary-max-value-default
git reset --hard origin/fix/color-to-binary-max-value-default
```

### What changed and why

The branch was conflicting with `main`, and the cause was worth explaining rather than just fixing.

Commit `22f6480` ("convert BGR/BGRA to grayscale in GrayToBinary") was byte-identical to what was already on `main`. Your own PR #349 landed that same fix and closed #319 about 26 minutes before you opened this PR. Because this branch was never rebased after that merge, git saw the same lines being introduced twice and flagged a conflict. I dropped that commit; the ColorToBinary fix is the only part still needed, and it now sits as a single commit.

The general habit worth picking up: `git fetch origin && git rebase origin/main` right before opening a PR, especially when you have another PR in flight touching nearby code. It would have caught this before review.

### What I kept from your work

Your rewritten `GrayToBinary` colour-input tests, which I had asked for in the first review round. They are genuinely better than the versions that reached `main` through #349: those use pixel values `128` and `200`, which both sit above the default threshold of `127`, so the output collapses to all-white and `issubset({0, 255})` passes without ever proving a binary split. Yours straddle the threshold and assert equality. My second review called them duplicates and asked you to drop them. That was wrong, and they are in the final commit.

I also strengthened your new ColorToBinary test from `result.max() > 0` to an equality check on `{0, 255}`, and mutation-checked it: with the old `maxValue` default restored it fails, so it genuinely pins the bug.

The commit is still authored to you.

### Two follow-ups, not for this PR

- `thresholdValue` still defaults to `0`, so default params now produce a mostly-white image rather than a useful binary split. Correct per #182 as written, but not a good default.
- `thresholding_applythreshold` (`thresholding.blocks.ts:6`) carries the identical `maxValue: 0` bug on a different block, with the same all-black result.

Either would be a reasonable thing to pick up next if you are interested.

PR #355, opened 2026-08-29
Test/backend integration and security tests

chetanbarakki:
## Description

Brief summary of the changes. Reference any related issues.
Added the following changes
- New exceptions for macro depth and memory limits.
- New limits enforced for macro recursion depth and image memory.
- Updated persistence logic for share link expiration.
- New test suites covering integration, step inspection, macro expansion, security tokens, and memory limits.

Fixes #39 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Existing tests pass
- [ ] New tests added
- [ ] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/imagelab-backend/app/exceptions.py
+++ b/imagelab-backend/app/exceptions.py
@@ -25,6 +25,20 @@ class PipelineExecutionError(Exception):
         super().__init__(user_friendly_message)
 
 
+class MacroDepthLimitExceeded(AppException, ValueError):
+    """Raised when macro expansion depth exceeds maximum allowed limit (10)."""
+
+    def __init__(self, message: str = "Macro expansion depth limit exceeded (> 10)"):
+        super().__init__(message=message, status_code=400)
+
+
+class MemoryLimitExceededException(AppException, ValueError):
+    """Raised when image processing memory usage exceeds configured threshold."""
+
+    def __init__(self, message: str = "Memory limit exceeded during image processing"):
+        super().__init__(message=message, status_code=400)
+
+
 async def app_exception_handler(request: Request, exc: AppException) -> JSONResponse:
     return JSONResponse(
         status_code=exc.status_code,
--- a/imagelab-backend/app/routers/persistence.py
+++ b/imagelab-backend/app/routers/persistence.py
@@ -171,7 +171,7 @@ def _get_active_share(session: Session, token: str) -> PipelineShare:
         if expires_at.tzinfo is None:
             expires_at = expires_at.replace(tzinfo=UTC)
         if expires_at < datetime.now(UTC):
-            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=INVALID_SHARE_DETAIL)
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Share token has expired")
 
     return share
 
--- a/imagelab-backend/app/services/graph_engine.py
+++ b/imagelab-backend/app/services/graph_engine.py
@@ -4,10 +4,13 @@ import uuid
 
 from sqlmodel import Session, select
 
+from app.exceptions import MacroDepthLimitExceeded
 from app.models.graph import GraphCycleError, GraphNode, PipelineGraph, topological_sort
 from app.models.persistence import PipelineVersion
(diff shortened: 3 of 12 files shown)


ivantha:
### Automated review summary

**Triage:** test  |  **Recommendation:** needs-revision

**Findings**
- **[major/security]** `imagelab-backend/app/services/pipeline_executor.py:103` check_memory_limit() is only wired into execute_pipeline's main step loop (line 227) and its pre-decode check (line 175). _run_sub_pipeline, which executes the steps inside macro_blend/macro_if_else branches, calls operator.compute() right here with no memory check at all. A single request whose pipeline contains a macro_blend/macro_if_else step with a branch step like resize_image(width=50000, height=50000) can allocate far past the 50MB threshold before any check fires, bypassing the new guard entirely. _(unverified)_
- **[major/security]** `imagelab-backend/app/routers/persistence.py:174` Expired share tokens now return 403 'Share token has expired', while the not-found branch just above (line 167) still returns 404 with the generic 'Invalid or expired share link'. This makes expired-vs-nonexistent tokens distinguishable again -- exactly what tests/test_persistence_api.py::test_share_obfuscation_and_expiry was written to prevent (per its name and inline 'obfuscation' comments). That test was edited in this PR to match the new behavior instead of catching the regression.
- **[major/bug]** `imagelab-backend/app/services/pipeline_executor.py:227` check_memory_limit(image) here runs inside the per-step try/except. Because MemoryLimitExceededException also subclasses ValueError, it gets caught by 'except ValueError' at line 254 and re-raised as PipelineExecutionError, which the router turns into HTTP 200 with success:false. The pre-execution check a few lines up (175-177) explicitly re-raises specifically to avoid this kind of swallowing and reaches the router's 400 path instead -- so the same exception produces two different HTTP contracts depending on whether the limit is hit before or during execution.
- **[minor/test]** `imagelab-backend/tests/test_memory_limits.py:40` Both memory-limit tests only cover an input image that is already oversized before any step runs (4500x4500x3 > 50MB at decode time). No test exercises a pipeline where a step inflates a small input past the limit -- the only scenario that reaches the inconsistent-handling path at pipeline_executor.py:227 or would surface the macro_blend/macro_if_else branch gap above.
- **[nit/style]** `imagelab-backend/app/exceptions.py:35` MacroDepthLimitExceeded and MemoryLimitExceededException are introduced side by side but use different naming conventions (Exceeded vs. ExceededException), and neither follows the existing PipelineExecutionError/GraphCycleError/GraphTypeError '...Error' suffix pattern used elsewhere in the codebase.

**Conventions**
- **PR title:** Branch-name-style slash ('Test/backend integration and security tests') instead of Conventional Commits syntax (type(scope): summary); the capitalized 'Test' also breaks convention.
- **Description:** Template placeholder text and every checklist box (including 'New tests added') were left unfilled; no rationale is given for the new limits or the share-link status-code change; and 'Fixes #39' would close a much broader tracking issue than this diff actually resolves.
- **Commits (1):** Sole commit 'Feat: BackendIntegration and security tests' capitalizes the type, runs two words together, and labels the change a feature when it's overwhelmingly new tests plus a couple of hardening/fix-sized changes (test:/fix: would fit better).
- _Align commit/PR titles with lowercase Conventional Commits syntax and actually fill in the PR template (tested-locally boxes, a short why) before merging._

This PR adds five new backend test files (macro-expansion depth, memory limits, persistence integration, share-token security, and step-inspection buffers) alongside two small hardening additions: a macro-expansion recursion cap and an image-memory ceiling. The memory guard only checks the main linear pipeline loop, though -- it's never invoked inside the macro_blend/macro_if_else branch sub-pipelines (_run_sub_pipeline), so a single request using an unbounded resize inside a branch can still blow past the intended cap. The share-link expiry change also quietly breaks the existing token-obfuscation guarantee (expired tokens now return a distinguishable 403 instead of the same 404 used for nonexistent ones), a regression the PR's own modified test doesn't catch because it was edited to match the new behavior rather than preserve the invariant. The test additions themselves are well-constructed and correctly exercise the existing endpoints; recommend closing the branch-execution gap and revisiting the share-link status code before merging.

PR #365, opened 2026-09-12
feat: Surface graph validation warnings in frontend before run (#197)

### LINKED ISSUE #197: [Feature] Surface graph validation warnings in the frontend before a run
**Difficulty:** Medium
**Area:** frontend

## Description
The backend half already landed. `_validate_graph` in `app/services/graph_engine.py` performs cycle detection plus per-port channel-type checks against `OPERATOR_PORT_SPECS`, raising `GraphTypeError` on a mismatch. It is only reachable from `compile_graph`, which runs inside the execution request, so a student still learns about an incompatibility only after a round trip to the server, as a raw error string.

What remains is the frontend half: a pre-run pass that surfaces those problems before the pipeline is sent. Today the only client-side check is the empty-graph guard in `Toolbar.tsx`, and `app/routers/pipeline.py` exposes no validate-only route, so an implementation either adds one that reuses `_validate_graph` or keeps a shared client-side copy of the channel contract. The result should be a modal listing readable warnings that offers both a fix path and a run-anyway path, so it guides without blocking experimentation.

## Acceptance Criteria
- [ ] Triggering a run performs a validation pass before the execution request is sent
- [ ] Problems are rendered in a modal as plain-language messages naming the offending blocks, not raw `GraphTypeError` text
- [ ] The modal offers both "run anyway" and a dismiss-and-edit path, so validation never hard-blocks a run
- [ ] A pipeline with no Read Image block is reported through the same modal rather than the current bare `setError` string
- [ ] Channel warnings agree with `OPERATOR_PORT_SPECS`, either by calling a validate-only backend route or by keeping one shared source of truth, so frontend and backend cannot drift
- [ ] A valid pipeline runs with no extra click and no modal
- [ ] Tests cover at least one incompatible pipeline and one clean pipeline

## Resources
- `imagelab-backend/app/services/graph_engine.py:243` - `_validate_graph`, the cycle and channel checks that already exist
- `imagelab-backend/app/services/graph_engine.py:292` - the only call site, inside `compile_graph`, during execution
- `imagelab-backend/app/services/graph_engine.py:18` - `OPERATOR_PORT_SPECS`, the per-operator channel contract to surface
- `imagelab-backend/app/routers/pipeline.py:26` - the only execution route; there is no validate-only endpoint yet
- `imagelab-frontend/src/components/Toolbar.tsx:216` - the current pre-run check, an empty-graph guard and nothing more
- `imagelab-frontend/src/hooks/usePipeline.ts:7` - `extractExecutableGraph`, which already builds the graph the check would inspect

<details>
<summary>Original report by @xxivani (2026-03-12)</summary>

Currently if a user connects blocks in an invalid order: for example feeding a grayscale operator's output into a colour only operator like BGR to HSV,  the error only surfaces as a cryptic runtime failure after the pipeline has already been sent to the backend. For a teaching platform this is a poor experience because students get no guidance on what went wrong or how to fix it.

A pre-run validation pass that inspects the pipeline before execution and surfaces known incompatibilities as human-readable warnings in a modal. The modal should give the user the choice to fix the issue or run anyway, so it never blocks experimentation.

Checks to include:

1. Grayscale operator fed into a colour-only operator (e.g. Gray Image → BGR to HSV)
2. Colour operator fed into a grayscale-only operator (e.g. Read Image → Apply Threshold without a Gray Image step)
3. No Read Image block present

Backend error messages after execution. These are harder to understand for students and only appear after a round-trip to the server.

**Additional context**
Any other context or screenshots.

</details>

shree1071:
Closes #197

## Summary

Adds a client-side validation pass that runs immediately when the user clicks Run. If the pipeline graph contains errors, a modal displays plain-language warnings before any request is sent to the backend. If the graph is valid, execution proceeds without any additional interaction.

### New files

- imagelab-frontend/src/utils/validatePipelineGraph.ts - Validation logic that mirrors the backend OPERATOR_PORT_SPECS and OPERATOR_OUTPUT_CHANNELS constants, performs a topological sort, and tracks channel state through the pipeline.
- imagelab-frontend/src/components/ValidationWarningModal.tsx - Modal component that presents the warning list and offers two actions: dismiss to edit, or proceed with execution.
- imagelab-frontend/tests/utils/validatePipelineGraph.test.ts - 17 unit tests covering valid pipelines, channel mismatches, missing Read Image blocks, control flow branches, and edge cases.

### Modified files

- imagelab-frontend/src/components/Toolbar.tsx - Calls validatePipelineGraph inside handleRun before dispatching the execution request.

## Validation checks

- Channel type mismatches (grayscale to color and color to grayscale)
- Missing Read Image block when no image input is present
- Invalid connections on mask ports and multi-input operators
- Missing required branches on macro_blend and macro_if_else nodes
- Recursive validation of nested control flow branches

## Acceptance criteria

- [x] A validation pass runs before each execution request
- [x] Warnings are rendered in a modal as plain-language messages that name the affected blocks
- [x] The modal offers both a dismiss path (return to editing) and a proceed path (run anyway)
- [x] A missing Read Image block is reported through the modal, not a bare setError call
- [x] Channel warnings are derived from the same operator specs as the backend
- [x] Valid pipelines execute immediately with no modal
- [x] Tests cover at least one incompatible pipeline and one clean pipeline

## Testing

Run the automated test suite:

npm run test validatePipelineGraph

17 tests pass.

## Implementation notes

Frontend validation mirrors backend rules to prevent drift. Cycle detection is intentionally left to the backend, as it is uncommon and more expensive to implement correctly on the client. Nested branches inherit the parent input channel rather than requiring a Read Image block of their own.

## Related

- Backend validation: \imagelab-backend/app/services/graph_engine.py\ (\_validate_graph\, line 243)
- Operator specs: \imagelab-backend/app/services/graph_engine.py\ (\OPERATOR_PORT_SPECS\, line 18)
- Execution entry point: \imagelab-frontend/src/components/Toolbar.tsx\ (\handleRun\)

### DIFF
--- a/imagelab-frontend/src/components/Toolbar.tsx
+++ b/imagelab-frontend/src/components/Toolbar.tsx
@@ -31,6 +31,8 @@ import VersionHistoryModal from "./VersionHistoryModal";
 import BatchProcessingModal from "./BatchProcessingModal";
 import CreateMacroModal from "./modals/CreateMacroModal";
 import ConfirmDialog from "./ConfirmDialog";
+import ValidationWarningModal from "./ValidationWarningModal";
+import { validatePipelineGraph, type ValidationWarning } from "../utils/validatePipelineGraph";
 
 /** Three-phase selection state for 2-click macro range picking. */
 type SelectionPhase = "idle" | "selecting" | "waitingForEnd";
@@ -82,6 +84,8 @@ export default function Toolbar({ workspace }: ToolbarProps) {
   const [showBatchModal, setShowBatchModal] = useState(false);
   const [showCreateMacroModal, setShowCreateMacroModal] = useState(false);
   const [showNewWorkspaceConfirm, setShowNewWorkspaceConfirm] = useState(false);
+  const [showValidationWarningModal, setShowValidationWarningModal] = useState(false);
+  const [validationWarnings, setValidationWarnings] = useState<ValidationWarning[]>([]);
 
   // ── Fallback single-click selection (used outside range-selection mode) ────
   const [selectedBlocks, setSelectedBlocks] = useState<Blockly.Block[]>([]);
@@ -222,6 +226,23 @@ export default function Toolbar({ workspace }: ToolbarProps) {
       return;
     }
 
+    // Validate pipeline graph before execution
+    const validationResult = validatePipelineGraph(graph, 3);
+
+    if (!validationResult.valid && validationResult.warnings.length > 0) {
+      // Show validation warning modal
+      setValidationWarnings(validationResult.warnings);
+      setShowValidationWarningModal(true);
+      return;
+    }
+
+    // Proceed with execution
+    executeGraphPipeline(graph);
+  };
+
+  const executeGraphPipeline = async (graph: ReturnType<typeof extractExecutableGraph>) => {
+    if (!originalImage) return;
+
     setExecuting(true);
     setError(null);
     setTiming(null);
@@ -271,6 +292,19 @@ export default function Toolbar({ workspace }: ToolbarProps) {
     } finally {
       setExecuting(false);
     }
+  };
+
+  const handleRunAnyway = () => {
+    setShowValidationWarningModal(false);
+    if (workspace) {
+      const graph = extractExecutableGraph(workspace);
+      executeGraphPipeline(graph);
+    }
+  };
+
+  const handleCancelValidation = () => {
+    setShowValidationWarningModal(false);
+    setValidationWarnings([]);
   };
 
   // Shift+? opens the shortcuts modal
@@ -603,6 +637,13 @@ export default function Toolbar({ workspace }: ToolbarProps) {
           }}
         />
       )}
+
+      <ValidationWarningModal
+        isOpen={showValidationWarningModal}
+        warnings={validationWarnings}
+        onRunAnyway={handleRunAnyway}
+        onCancel={handleCancelValidation}
+      />
 
       <ConfirmDialog
         isOpen={showNewWorkspaceConfirm}
--- a/imagelab-frontend/src/components/ValidationWarningModal.tsx
+++ b/imagelab-frontend/src/components/ValidationWarningModal.tsx
@@ -1,0 +1,144 @@
+import { useEffect, useRef } from "react";
+import { AlertTriangle, X } from "lucide-react";
+import type { ValidationWarning } from "../utils/validatePipelineGraph";
+
+interface ValidationWarningModalProps {
+  isOpen: boolean;
+  warnings: ValidationWarning[];
+  onRunAnyway: () => void;
+  onCancel: () => void;
+}
+
+export default function ValidationWarningModal({
+  isOpen,
+  warnings,
+  onRunAnyway,
+  onCancel,
+}: ValidationWarningModalProps) {
+  const runAnywayButtonRef = useRef<HTMLButtonElement>(null);
+
+  useEffect(() => {
+    if (isOpen) {
+      // Focus the run anyway button when dialog opens
+      runAnywayButtonRef.current?.focus();
+    }
+  }, [isOpen]);
+
+  useEffect(() => {
+    if (!isOpen) return;
+
+    const handler = (e: KeyboardEvent) => {
+      if (e.key === "Escape") {
+        onCancel();
+      }
+    };
+    document.addEventListener("keydown", handler);
+    return () => document.removeEventListener("keydown", handler);
+  }, [isOpen, onCancel]);
+
+  if (!isOpen) return null;
+
+  return (
+    <div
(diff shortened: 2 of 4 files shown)


ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-revision

**Findings**
- **[major/bug]** `imagelab-frontend/src/utils/validatePipelineGraph.ts:151` The hasReadImage check tests node.type against "basic_input", "readimage", or "read_image", but the app's only registered Read Image block type is "basic_readimage" (imagelab-frontend/src/blocks/definitions/basic.blocks.ts:9). extractExecutableGraph roots the graph at that block (extractWorkspaceGraph in imagelab-frontend/src/utils/extractMacroGraph.ts:375) and keeps it as a node typed "basic_readimage", so the check never matches it; and a workspace with no Read Image block yields an empty graph, which Toolbar.tsx:224 rejects before validation runs. So hasReadImage is false on every real pipeline, and the "No Read Image block found" modal appears on every Run click, forcing "Run anyway" each time. Since the empty-graph case is already handled upstream, this check can probably be dropped rather than fixed.
- **[minor/test]** `imagelab-frontend/tests/utils/validatePipelineGraph.test.ts:32` The test suite's Read Image fixtures use the same incorrect type strings ("basic_input", "readimage", "read_image") as the implementation instead of the app's real block type "basic_readimage", so the 17 tests validate the mirrored logic against itself rather than against graphs shaped like what extractExecutableGraph actually produces, which is why they pass despite the hasReadImage bug above.
- **[minor/quality]** `imagelab-frontend/src/utils/validatePipelineGraph.ts:249` The try/catch around the topological sort is effectively dead for its stated purpose: topologicalSort (Kahn's algorithm, lines 165-213) never throws when the graph has a cycle, it simply omits the cycled nodes from the returned order, so those nodes are skipped in the validation loop with no warning surfaced at all. The comment "If topological sort fails (cycle), that's handled by backend" is misleading since no exception is thrown for that case; the catch only fires for a different failure (e.g. an edge referencing a node id not present in the graph).

**Conventions**

The PR adds a well-documented client-side pre-execution validator and warning modal, but the Read Image detection is broken: it checks for node types "basic_input"/"readimage"/"read_image", while the app's only Read Image block type is "basic_readimage", which is the root of every executable graph. As shipped, the "No Read Image block found" warning surfaces on every pipeline run, directly contradicting the PR's own acceptance criterion that valid pipelines execute without a modal. The 17 new unit tests encode the same incorrect type strings as the implementation, so they pass without catching this. A secondary, lower-priority issue: the try/catch around the topological sort is dead code for the cycle case it claims to handle, so cyclic graphs silently skip type-checking with no warning shown to the user. Commit hygiene and the PR description are otherwise clean and informative.

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-revision

**Findings**
- **[major/quality]** `imagelab-frontend/src/components/ValidationWarningModal.tsx:108` ValidationWarningModal autofocuses the "Run Anyway" button when the dialog opens (runAnywayButtonRef.current?.focus() in the isOpen effect), so a user who presses Enter/Space reflexively after the modal pops up — a very plausible reaction to a warning dialog — executes the pipeline despite the warnings instead of dismissing it. The safer default is to focus the "Cancel and Fix" button, or no button at all, given this is a warning dialog meant to make the user pause.
- **[minor/quality]** `imagelab-frontend/src/components/Toolbar.tsx:297` handleRunAnyway re-derives the graph from the live workspace (extractExecutableGraph(workspace)) rather than executing the graph object that was actually validated and shown to the user. If the workspace changes between the warning modal opening and the user clicking "Run Anyway", the executed graph could silently differ from the one whose warnings were reviewed.
- **[nit/style]** `imagelab-frontend/src/components/ValidationWarningModal.tsx:177` The warnings list is keyed by array index (key={index}) rather than a stable identifier such as `${warning.nodeId}-${warning.port}-${index}`. Harmless today since the list is never reordered or mutated in place, but it's a recognized React anti-pattern worth avoiding.

**Conventions**
- **Commits (2):** clean, conventional history
- _Clean history for the size of the change: one feature commit plus one focused fixup commit (correcting the Read Image block type check and adding cycle detection) before merge._

This PR adds a client-side validation pass (channel-type checks, missing-Read-Image detection, control-flow branch checks, and cycle detection) that mirrors the backend's graph_engine rules, surfacing plain-language warnings in a modal before a pipeline runs. The logic is well-tested (17 unit tests covering the main branches) and the PR description is thorough and accurately maps to the diff. The main actionable issue is that the warning modal autofocuses "Run Anyway" rather than the safer "Cancel and Fix" action, which undercuts the feature's purpose by making it easy to blow past a warning with a stray Enter keypress — worth fixing before merge. The other two findings are minor robustness/style nits.

